PR #1, opened 2026-07-19
Optional token auth, opt-in autonomy/gate controls, and a security test suite

bergmorten:
## What

Defense-in-depth for the server's capability surface (shell / git write / docker / chat), plus a test suite for the security-critical parsers. **Zero new dependencies** (stdlib Python / Bun / node built-ins only), and the **default local single-user experience is unchanged** — every new control is opt-in via env.

The server today rests entirely on *loopback + same-origin*, which is documented and fine for a single-user box but leaves two gaps: any other local process can reach the port (and the shell), and a non-loopback bind is unauthenticated RCE. This adds a way to close both without changing the zero-config path.

## Changes

| Env / area | Behavior |
|---|---|
| `AGENTGLASS_TOKEN` | Shared secret required on every route **except** the append-only telemetry sinks (`/ingest`, OTLP). `Authorization: Bearer` or `?token=` (WS/downloads). Binding a non-loopback address **without** a token now refuses to run unauthenticated: it mints, prints, and `0600`-saves one. |
| `AGENTGLASS_CHAT_BYPASS` | `bypassPermissions` (`claude --dangerously-skip-permissions`) honored only when opted in; otherwise downgraded to a prompting default. |
| `AGENTGLASS_GATE_FAILCLOSED` | Gate timeout (server) / unreachable control plane (hook) **denies** instead of allows. Off by default. |
| `AGENTGLASS_RATE_MAX` / `_WINDOW_MS` | Fixed-window limiter on the unauthenticated intake sinks (flood protection). |
| CORS | Reflects the caller's `Origin` (+ `Vary: Origin`) instead of a blanket `*`; permits the `Authorization` header. |
| Tests | `server/test/security.test.ts` — 16 cases over `privateHost` (incl. the `10.evil.com` rebinding case), `safeAbs`, `shellSafeRel`, `parseMakeTargets` (incl. the `-flib/*.mk` make-flag injection), and the token check. |
| Typecheck | `Bun.serve<WsData>` resolves the WS-data generic — `server/src` now passes `tsc --noEmit` clean (was 11 errors), so it could join a typecheck gate. |

Web side: the token is read once from `?token=` (then stripped from the address bar) or `localStorage` and attached to every fetch/WS/download.

## Verification

- `bun test` → 16/16 pass
- `bunx tsc --noEmit` → clean in both `web/` **and** `server/`
- `bunx vite build` → ok
- Boot-tested: default (loopback, no token), `AGENTGLASS_TOKEN` (401 without / 200 with, intake ungated), and `AGENTGLASS_BIND=0.0.0.0` with no token (auto-mints + prints).

## Two items I'd like your call on (kept separate in the diff so they're easy to drop)

1. **`AGENTGLASS_TRUST_LAN` default = off** (commit 1). Today RFC1918 addresses are trusted as origins/hosts; this makes the default loopback-only, so a machine currently accessed from a LAN browser would need `AGENTGLASS_TRUST_LAN=1`. It's the one **behavior change** here. Happy to flip the default to on (preserving today's behavior) and let the token be the real gate — your call, since the RFC1918 trust was a deliberate design choice.
2. **`tauri.conf.json` `targets: "all"`** (commit 3). This contradicts the "Linux only, `.deb`" note in `CONTRIBUTING.md` and could change `make desktop` on a minimal Linux box (now also attempts rpm/appimage). Easy to revert to `["deb"]` if you'd rather add other platforms yourself.

Everything else is additive and default-off. Glad to split, trim, or reshape to fit.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,23 +283,35 @@ write to your repos and control Docker. It ships safe for its intended home —
 **your own single-user machine** — and you should know exactly where the lines
 are:
 
-- **It only listens on your own machine.** The server binds `127.0.0.1` — 
-  nothing on your network can reach it. There is **no authentication**, because
-  on a single-user machine "can reach localhost" already means "is you".
+- **It only listens on your own machine.** The server binds `127.0.0.1` —
+  nothing on your network can reach it. By default there is **no authentication**,
+  because on a single-user machine "can reach localhost" already means "is you".
+- **Optional shared-secret token.** Set `AGENTGLASS_TOKEN` and every route but
+  the append-only telemetry sinks (`/ingest`, the OTLP receivers) requires it —
+  `Authorization: Bearer <token>` for the API, `?token=<token>` on the dashboard
+  URL (it's stored and stripped from the address bar). This is what makes a
+  shared machine or a network bind safe, and it stops *other local processes*
+  from opening the shell. Binding a non-loopback address **without** a token
+  refuses to run unauthenticated: it mints one, prints it, and saves it
+  `0600` under your config dir.
 - **Websites you visit can't touch it.** Every request is origin-checked, the
   shell and the live stream require a verified local origin, and a Host-header
   guard blocks DNS-rebinding tricks (browsers can't forge `Host`). Running it
   behind a reverse proxy? Allow its name via `AGENTGLASS_ALLOWED_HOSTS`.
-- **⚠️ Shared / multi-user machines are NOT the intended home.** `localhost`
+- **⚠️ Shared / multi-user machines are NOT the default home.** `localhost`
   belongs to the *machine*, not to your account — on a box where other people
   also have accounts, any of them could reach the server and its shell **as
-  your user**. If you must run it there, disable the capability surfaces:
-  `AGENTGLASS_TERMINAL_DISABLED=1`, `AGENTGLASS_CHAT_DISABLED=1`,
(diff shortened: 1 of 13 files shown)


SirAllap:
Merged as-is — thank you, this is a fantastic contribution. I pulled the branch and verified everything before merging: 16/16 tests, clean typecheck on both sides, build ok, and the token flow behaves exactly as described (401 without, 200 with Bearer or ?token=, intake open, and the PTY shell properly behind the token). The auto-mint-on-exposed-bind is a genuinely nice touch, and you gave this repo its first test suite, which says something.

On your two questions, rather than make you push another round I merged and handled them in a follow-up commit:

1. **TRUST_LAN default off: kept your default.** It only affects already-exposed setups, and it makes the "exposing is a deliberate act" line in the README literally true. Good call.
2. **tauri targets: flipped back to `["deb"]`** — CONTRIBUTING still says Linux only, and I'd rather add platforms deliberately when someone can test them. Nothing wrong with your change, just sequencing.

One small thing I added on top: `/chat/enabled` now carries the bypass opt-in flag so the chat mode picker stops offering `bypassPermissions` when the server would silently downgrade it — your downgrade stays as the real enforcement, the UI just stops lying about it.

Hope to see you around here again — contributions like this one are exactly what this project needs.

---

PR #23, opened 2026-07-19
fix(hooks): quote send_event path on Windows installs

### LINKED ISSUE #15: Windows hook installation fails due to unquoted paths
### Issue

On Windows, running `python3 hooks/install_hooks.py` generates hook commands with unquoted paths, which breaks the Claude Code settings when the path contains spaces or backslashes.

### Error

When `install_hooks.py` runs, it generates hook commands like:
```
python3 C:\Users\Guillaume\Downloads\Scripts\agentglass\hooks\send_event.py --event-type SessionStart
```

This fails because:
1. Windows paths with backslashes aren't properly escaped in JSON
2. Spaces in the path break the command parsing
3. Claude Code then fails to execute the hook, breaking the entire installation

### Result

- Claude Code becomes unusable with error messages in every tool call
- Settings file gets corrupted and needs manual restoration
- Installation fails silently for users on Windows

### Solution

Quote the path on Windows. In `hooks/install_hooks.py` line 72, change:
```python
cmd = f"python3 {SEND_EVENT} --event-type {event}"
```

To:
```python
send_event_cmd = f'"{SEND_EVENT}"' if os.name == 'nt' else SEND_EVENT
cmd = f"python3 {send_event_cmd} --event-type {event}"
```

This generates the correct format:
```
python3 "C:\Users\Guillaume\Downloads\Scripts\agentglass\hooks\send_event.py" --event-type SessionStart
```

### Environment

- OS: Windows 11
- Python: 3.12
- Bun: 1.3.14

Solaris-star:
## Summary
Fixes #15

On Windows, `hooks/install_hooks.py` wrote:

```text
python3 C:\Users\...\hooks\send_event.py --event-type SessionStart
```

Unquoted paths with spaces/backslashes break Claude Code settings parsing and can make every tool call fail.

### Change
Quote the script path when `os.name == "nt"`:

```text
python3 "C:\Users\...\hooks\send_event.py" --event-type SessionStart
```

Non-Windows installs stay unchanged.

## Testing
- Inspected the single command construction site in `hooks/install_hooks.py`
- Confirmed quote branch is Windows-only

### DIFF
--- a/hooks/install_hooks.py
+++ b/hooks/install_hooks.py
@@ -69,7 +69,9 @@ def do_install(cfg):
     hooks = cfg.setdefault("hooks", {})
     for event, (matcher, add_chat) in EVENTS.items():
         arr = [e for e in hooks.get(event, []) if not _is_ours(e)]
-        cmd = f"python3 {SEND_EVENT} --event-type {event}"
+        # Quote the script path on Windows so spaces/backslashes survive JSON + shell parsing.
+        send_event_cmd = f'"{SEND_EVENT}"' if os.name == "nt" else SEND_EVENT
+        cmd = f"python3 {send_event_cmd} --event-type {event}"
         if add_chat:
             cmd += " --add-chat"
         entry = {"hooks": [{"type": "command", "command": cmd}]}


SirAllap:
Merged, thank you! Small and exactly right.

One thing I noticed while looking at it: paths with spaces break the same way on macOS and Linux, so I'm going to drop the os.name check in a follow-up and just quote the path everywhere. Your fix stands as-is for Windows.

Really appreciate you picking up someone else's bug report and fixing it, that's the good stuff. Hope to see you around.

---

PR #28, opened 2026-07-19
fix(hooks): pick Windows python launcher for hook commands

### LINKED ISSUE #24: Windows: hook commands invoke python3, which usually doesn't exist there
Split out of #15 (the quoting half was fixed by #23). The hook commands written by `hooks/install_hooks.py` start with `python3`:

```
python3 "C:\...\hooks\send_event.py" --event-type SessionStart
```

On most Windows installs there is no `python3` command — the launcher is `py` and the interpreter is `python` (a `python3` alias only exists via the Microsoft Store shim). So even with the path properly quoted, every hook likely fails with `'python3' is not recognized`, which Claude Code surfaces on every tool call.

## Fix direction

Have the installer pick the interpreter per platform when writing the command — e.g. on `os.name == "nt"` prefer `py` (falling back to `python`), keep `python3` elsewhere. Deliberately NOT `sys.executable`: it pins an absolute interpreter path that breaks when a venv moves or is deleted.

Same consideration applies to `hooks/connect_otel.py` docs and the `gate_event.py` examples in the README.

## Verification wanted

I don't have a Windows machine on hand — if you run agentglass hooks on Windows (@JGtm reported the original issue from Win 11), a confirmation of what `python3` resolves to there, and whether `py`-prefixed commands work end to end, would let us close this for real.

Solaris-star:
## Summary
On Windows, Claude Code hook commands generated by `hooks/install_hooks.py` used `python3`, which is usually missing (Microsoft Store shim / not installed). Prefer `py` then `python` on Windows; keep `python3` elsewhere.

Also notes the same Windows launcher choice in `settings.example.json` and the README gate example. Path quoting remains unconditional on all platforms (already merged via #23).

Fixes #24

## Notes
- Avoids `sys.executable` so a deleted venv cannot brick every hook
- Re-running the installer rewrites existing agentglass hook commands in place (idempotent strip+append)

## Testing
- `python3 -m py_compile hooks/install_hooks.py`
- Manual inspection of `_hook_python()` branch selection

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,6 +330,8 @@ with Approve / Deny buttons — decide from any device and the agent unblocks.
 "PreToolUse": [
   { "matcher": "Bash", "hooks": [{ "type": "command",
     "command": "python3 ~/code/agentglass/hooks/gate_event.py --source-app my-project" }] }
+
+On Windows, use `py` (or `python`) instead of `python3` in hand-written hook commands; `hooks/install_hooks.py` picks this automatically.
 ]
 ```
 
--- a/README.md
+++ b/README.md
@@ -330,10 +330,10 @@ with Approve / Deny buttons — decide from any device and the agent unblocks.
 "PreToolUse": [
   { "matcher": "Bash", "hooks": [{ "type": "command",
     "command": "python3 ~/code/agentglass/hooks/gate_event.py --source-app my-project" }] }
-
-On Windows, use `py` (or `python`) instead of `python3` in hand-written hook commands; `hooks/install_hooks.py` picks this automatically.
 ]
 ```
+
+On Windows, use `py` (or `python`) instead of `python3` in hand-written hook commands; `hooks/install_hooks.py` picks this automatically.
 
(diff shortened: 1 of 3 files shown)


SirAllap:
This is great — the interpreter detection is exactly what #24 needed, and avoiding sys.executable for the venv reason tells me you actually read the issue. Thank you for coming back for a second one!

One tiny thing I spotted: the README note landed inside the JSON code block (right before the closing `]` of the gate example), so the example renders with the sentence mixed into the JSON. Moving it just below the closing ``` fence fixes it.

Totally happy to do that tweak myself after merging if you're busy — just say the word. Either way this is going in, and I'll ping the original reporter afterwards to confirm on a real Windows machine.

---

Solaris-star:
Fixed — moved the Windows `py`/`python` note to just below the closing fence so the JSON example stays valid. Thanks for the catch!

SirAllap:
Merged! Really appreciate you turning the README tweak around that fast. Two solid fixes from you in one day — the Windows folks owe you one 🙌

---

Issue #10, opened 2026-07-19
Tests: grow coverage beyond the security suite (ingest, pricing, parsers, derive)


SirAllap:
#1 added `server/test/security.test.ts` (16 tests) — great foundation, but the data path that everything else sits on is still untested. These are mostly **pure functions**, so tests are cheap to write and each one is a self-contained contribution:

## Wishlist (each bullet is a fine standalone PR)

- **`ingest.ts` / `normalize`** — hook payloads → events: token extraction, cumulative→delta accounting (the "how cost stays correct" invariant in the README), `detectError` on tool_response shapes
- **`pricing.ts`** — model-name substring matching (does `claude-sonnet-5` hit the right row?), cache-read/write rates, unknown models → $0 not NaN
- **`otlp.ts`** — a recorded `gen_ai.*` span batch → expected events (fixtures from Gemini CLI / Codex CLI make this real)
- **`terminal.ts` / `makeCommands` + `scriptCommands`** — the Makefile description conventions (`## inline`, `# above`), `::=` exclusion, runner detection per lockfile
- **`web/src/lib/derive.ts`** — agent status derivation (running/idle/stuck) from an event buffer, alert rules

## Ground rules

`bun test`, no new deps, fixtures inline or under `server/test/fixtures/`. Once #5 (CI) lands, these run on every PR automatically.

SirAllap:
Progress update — three wishlist items landed today, all from @Solaris-star:

- [x] `pricing.ts` → #31 (model matching, cache rates, unknown fallback)
- [x] `ingest.ts` `detectError` + transcript token sums → #32
- [x] `terminal.ts` `makeCommands` / `scriptCommands` → #33

Still open for anyone who wants a self-contained first PR:

- [ ] `ingest.ts` / `normalize` — cumulative→delta token accounting (the cost-correctness invariant)
- [ ] `otlp.ts` — recorded `gen_ai.*` span batch → expected events
- [ ] `web/src/lib/derive.ts` — status derivation (running/idle/stuck) + alert rules

Suite is at 38 tests across 4 files and runs on every PR via CI.

---
_Generated by [Claude Code](https://claude.ai/code)_

SirAllap:
Closing this as done.

When it was written the suite was `server/test/security.test.ts` and 16 tests. The four bullets on the wishlist all have coverage on `main` now:

[...]

PR #32, opened 2026-07-19
test(ingest): cover detectError shapes and transcript token sums

Solaris-star:
## Summary
Adds pure-function tests for `detectError` and `sumTranscriptTokens` from the #10 coverage wishlist.

- PostToolUseFailure / is_error flags
- tool_response success:false, interrupted, returnCodeInterpretation
- strong shell-failure markers
- successful / benign stdout not false-positive
- transcript usage summing + prompt_tokens aliases

Part of #10

## Testing
```
cd server && bun test test/ingest.test.ts
# 9 pass
```


SirAllap:
Merged. The "benign stdout mentioning error" case is a great catch — that false positive is the one I was most worried about in detectError. Thank you!

---

PR #33, opened 2026-07-19
test(terminal): cover makeCommands and scriptCommands discovery

Solaris-star:
## Summary
Adds tests for `makeCommands` / `scriptCommands` / `shellSafeRel` from the #10 coverage wishlist.

- Makefile `##` inline and comment-above descriptions
- nested `make -C <dir>` targets
- skip leading-dash fake targets
- bun.lock → `bun run --cwd`
- filter unsafe script names

Part of #10

## Testing
```
cd server && bun test test/terminal-commands.test.ts
# 4 pass
```


SirAllap:
Merged! And with that, all three of today's test PRs are in — I ran them together on top of main before merging: 38 pass across 4 files. Real temp dirs for the make -C paths was the right call, way more honest than mocking.

Seriously, thank you for this batch. #10's wishlist is shrinking fast because of you 🙌

---

PR #58, opened 2026-07-19
fix(server): attribute Windows-side transcripts to the right projects under WSL

icey2488:
## What does this PR do?

Hi! You mentioned being interested in feedback from Windows/WSL setups — here's a fix for an attribution bug I hit running agentglass under WSL2 while scanning Windows-side Claude Code transcripts.

### Problem

Running the server in WSL2 (Ubuntu 24.04) with `AGENTGLASS_PROJECTS_DIR` pointed at the Windows-side `~/.claude/projects` (via `/mnt/c`), every session collapsed into a single project — the agentglass repo itself. The dashboard showed 1 project when there should have been ~11.

### Root cause

Windows-side Claude Code records Windows cwds (`C:\Users\...`) in its transcripts. On a posix host that string isn't an absolute path, so `resolve()` in `safeAbs` treats it as *relative to the server's own cwd*. `projectRootOf` then hands that phantom path to `git rev-parse`, which walks up and finds the repo the server runs from — so every Windows session gets attributed to agentglass.

### Fix

- `server/src/git.ts`: in `safeAbs`, detect drive-letter paths (`^[A-Za-z]:[\\/]`) and map them onto the WSL automount (`/mnt/c/...` by default, honoring a custom `[automount] root` in `/etc/wsl.conf`) before any `resolve()`/git logic sees them. No-op on native Windows (`process.platform === "win32"`) and for paths that are already posix.
- `server/src/transcripts.ts`: when a cwd has no resolvable repo, fall back to `safeAbs(cwd)` instead of the raw cwd, so a translated Windows cwd still groups under its own folder rather than a raw `C:\...` string. This matters in practice because git often refuses `/mnt/c` repos with `detected dubious ownership`, so the fallback is what labels those projects.

## How was it tested?

Environment: WSL2 Ubuntu 24.04, bun, transcripts on `/mnt/c` over 9p.

Startup scan line, before → after (same machine, same transcript tree):

```
📚 scanned /mnt/c/Users/Raide/.claude/projects — 27086 events from 1 project in 11385ms   (main)
📚 scanned /mnt/c/Users/Raide/.claude/projects — 27090 events from 10 projects in 12790ms  (this branch)
```

- Totals are preserved: 345 sessions in both runs; the tiny event delta is live drift from an active session appending during the runs. `/projects` now lists distinct, sensible entries (`claunker-hermes`, `worldclock-discord`, `kanbantt-app`, …) instead of one.
- Perf: the added git probes are already bounded by the existing per-cwd `rootCache` memoization (one probe per distinct cwd), so a full ~27k-event rebuild only went from ~11.4s to ~12.8s, and steady-state incremental sweeps are unchanged (~1–3s here).
- `bunx tsc --noEmit` clean; `bun test` 16/16 pass.

---

Happy to adjust anything — naming, where the translation lives, or splitting the fallback change out. Thanks for the project, it's been genuinely useful for watching agents across a whole machine.


SirAllap:
Thank you for this — and for taking the time to test on a setup I don't have and can't easily reproduce.

The writeup did most of the work for me. `resolve()` treating `C:\Users\...` as *relative* on a posix host, and `git rev-parse` then walking up into whatever repo the server happens to run from, is something I'd have spent a long evening chasing from the symptom alone ("dashboard shows 1 project"). Reading `[automount] root` out of `/etc/wsl.conf` rather than hardcoding `/mnt/` is a detail I wouldn't have known to handle, and I appreciate you explaining *why* the extra git probes stay cheap (the per-cwd `rootCache` bounding them) instead of just asserting it.

I pulled the branch and checked it on top of current `main` — which moved a fair bit today, so this is the merged state rather than your original base:

- `bun test` — 79 pass across 9 files
- `bunx tsc --noEmit` clean in both `server/` and `web/`
- `bunx vite build` clean
- `safeAbs` spot-checks: `C:\Users\Raide\code\app` → `/mnt/c/Users/Raide/code/app`, `D:/projects/thing` → `/mnt/d/projects/thing`, posix paths untouched, relative paths unchanged

**One thing I wanted to run past you**, and you may well have already thought about it. Translating `\` → `/` also turns `..` into real traversal, which it wasn't before — on posix a backslash isn't a separator, so `C:\..\..\etc\passwd` used to collapse into one literal segment under the cwd and couldn't climb anywhere. After the change it resolves out of the automount:

```
C:\..\..\..\etc\passwd    →  /etc/passwd
C:\..\..\home\me\.ssh     →  /home/me/.ssh
```

I don't think it's exploitable as things stand — `repoRootOf`/`guard()` still require a real repo root downstream — so this is belt-and-braces rather than a live hole. But `safeAbs` also sees request-supplied paths (git ops, terminal cwd, chat cwd), not only transcript data, so it felt worth mentioning before it goes in.

If it seems worth guarding, something like clamping back under the mount after resolving would do it:

```ts
const abs = resolve(fromWindowsPath(p));
// A translated drive path should stay inside the mount it maps to: `\` became
// a real separator, so `..` in a Windows-recorded cwd can now climb out.
if (drivePath && !(abs === mountRoot || abs.startsWith(mountRoot + "/"))) return null;
```

Happy for you to add it, happy to push it as a follow-up myself, and equally happy to be told I've misread the risk — you've clearly spent more time in WSL path semantics than I have. It's your fix either way and your name should stay on it.

This is going in regardless; the attribution bug is real and 1 → 10 projects on your machine speaks for itself. Thanks again, and for offering to split or rename anything — that made reviewing it easy.

---

icey2488:
You're right, and thanks for catching it — I'd been thinking of safeAbs as transcript-only, but since it also sanitizes request-supplied paths, the translation making `..` traversable is exactly where fail-closed belongs. Pushed the clamp on top: a translated drive path must resolve to its own mount root or below, else null — plus tests for the escape case (`C:\..\..\etc\passwd` → null) and legitimate inner `..` (stays contained); POSIX passthrough is unchanged (the existing tests still pin it). One deliberate non-change: UNC paths stay untranslated, so they keep the old collapse-to-literal containment. Happy for this to land whenever — and thanks for the kind of review that makes a first PR easy.

SirAllap:
Thank you so much for this — genuinely one of the most careful PRs I've seen on the project. 🙏

The drive-letter → WSL automount mapping is exactly the right layer to fix it, and I really appreciate that you thought through the security side too — clamping so `..` can't climb out of the mount or hop to a sibling drive is a lovely touch, and the fact you covered all of it with tests (plus the UNC case) made this a joy to review. The before/after scan output and the perf note were the cherry on top.

This reads great to me and I'm glad it builds on the `resolvedRoot` cache without any friction. Really happy agentglass has been useful for watching agents across a whole machine — thanks for taking the time to send such a thorough fix. 🙌

---

SirAllap:
Merged! 🎉 Thanks again for such a careful, well-tested fix — WSL folks are going to be glad you hit this before they did. Really appreciate you contributing. 🙌

---

PR #70, opened 2026-07-19
fix(stats): stop ?window=all zeroing every total

icey2488:
### Problem

`GET /stats?window=all` answers with all-zero totals:

```json
{"totals":{"events":0,"sessions":0,...},"window_ms":null}
```

### Root cause

In `server/src/index.ts`:

```ts
const windowMs = Math.min(3660 * 86_400_000, Math.max(60_000, Number(url.searchParams.get("window") || 24 * 3600 * 1000)));
```

`Number("all")` is `NaN`, and `Math.min`/`Math.max` pass NaN straight through, so `statsSummary` computes `since = Date.now() - NaN` and every `timestamp > since` comparison matches nothing — zero totals, `window_ms: null` in the JSON. The dashboard never trips over this because its "all" chip sends `3650 * 86_400_000` as a number; only direct API calls hit it.

### Fix

The param now parses through a small `parseWindowMs` helper (its own module, so tests can import it without booting the server):

- `"all"` (any case) → the existing 3660-day cap — same order of magnitude as the UI chip, so both spellings of "everything" cover the whole database;
- numbers → clamped to [1 minute, cap] exactly as before;
- absent → the 24h default, unchanged;
- anything malformed → the 24h default rather than NaN.

One judgment call: malformed input falls back to the default rather than to "all" — a typo silently returning months of data seemed more surprising than getting the default window. Easy to flip if you'd rather.

## How was it tested?

- New `server/test/params.test.ts` covers "all"/case, clamping, garbage, and that every path returns a finite number; `bun test` green (101 pass), `bunx tsc --noEmit` clean.
- Live before/after on my install (~27.5k events over 66 days):
  - before: `window=all` → `events: 0`, `window_ms: null`
  - after: `window=all` → `events: 27535, sessions: 351`, `window_ms: 316224000000`
  - numeric windows unchanged (`window=3600000` → 404 events); `window=banana` → 24h default.

Environment: WSL2 Ubuntu 24.04, scanning a Windows-side `~/.claude/projects` over `/mnt/c`.

---

This is one of three small PRs from the same install (with `fix/uptime-stat` and `feat/multi-root-projects-dir`); they're independent and any subset merges cleanly.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -47,6 +47,7 @@ import { workspaceRoot, setWorkspaceRoot, CONFIG_PATH } from "./config.ts";
 import { privateHost } from "./net.ts";
 import { resolveToken, tokenOk, isIntake, isAuthExempt } from "./auth.ts";
 import { rateOk } from "./ratelimit.ts";
+import { parseWindowMs } from "./params.ts";
 
 const PORT = Number(process.env.AGENTGLASS_PORT || 4000);
 /**
@@ -521,7 +522,7 @@ const server = Bun.serve<WsData>({
       return json(getSessions(limit, url.searchParams.get("provider") || undefined));
     }
     if (pathname === "/stats") {
-      const windowMs = Math.min(3660 * 86_400_000, Math.max(60_000, Number(url.searchParams.get("window") || 24 * 3600 * 1000)));
+      const windowMs = parseWindowMs(url.searchParams.get("window"));
       return json(statsSummary(windowMs, url.searchParams.get("provider") || undefined));
     }
 
--- a/server/src/params.ts
+++ b/server/src/params.ts
@@ -1,0 +1,23 @@
+// Query-param parsing kept out of index.ts so tests can import it without
+// booting the server.
+
+/** Default /stats window: the last 24 hours. */
+export const DEFAULT_WINDOW_MS = 24 * 3600 * 1000;
+/** Widest window we compute over — same order of magnitude as the UI's "all"
+ *  chip (3650d), so both spellings of "everything" cover the whole database. */
+export const MAX_WINDOW_MS = 3660 * 86_400_000;
+
+/**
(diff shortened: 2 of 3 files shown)


SirAllap:
Merged #69 and #71 — thank you, all three were a pleasure to review.

I verified this one on my own machine before touching it, and both bugs reproduced exactly as you described:

```
/stats?window=all  →  events: 0 · window_ms: None
```

`Number("all")` → `NaN` sailing through `Math.min`/`Math.max` is a lovely find, and the reason it never showed up in the UI (the "all" chip sends a number) is exactly the kind of detail that makes a report easy to act on. After your fix, against a copy of my real database: `events: 7517`, `window_ms: 316224000000`, and `window=banana` → the 24h default. Numeric windows unchanged.

Your judgment call on malformed input falling back to the default rather than to "all" is the right one, for the reason you gave — a typo quietly returning months of data is the more surprising failure.

**One thing I should flag, and it's my doing rather than yours.** The three PRs are independent against `main` as it stood, but #69 and this one both rewrite the same `/stats` line, so once #69 landed this went to CONFLICTING. Since "allow edits by maintainers" was on, I pushed the resolution to your branch rather than making you round-trip for it — I hope that's welcome, and tell me if you'd rather I hadn't. The merge keeps both intentions:

```ts
const windowMs = parseWindowMs(url.searchParams.get("window"));
return json({ ...statsSummary(windowMs, ...), server_started_at: STARTED_AT });
```

Nothing of yours was changed apart from that line. `bun test` 128 pass, `tsc --noEmit` clean in both packages on the merged result.

Also worth saying: your CI hadn't run at all because fork workflows need maintainer approval and I hadn't noticed the three runs sitting in the queue. That was on me — approved now, and all three are green.

Four solid PRs from a setup I can't reproduce, each with the root cause traced and the fix explained. Thank you — the WSL side of this project is in much better shape than it was this morning.

---

Issue #46, opened 2026-07-19
Make the radar legible: the "distance = context-to-compaction" encoding isn't obvious


SirAllap:
## The problem

The radar encodes something genuinely useful — **distance from centre = how full the context window is**, so a blip drifting to the edge is about to compact. But that meaning isn't legible at a glance: to a first-time viewer it reads as decorative sci-fi, not a gauge. A visualization whose meaning you have to be told is doing less work than it could.

## Current state
`web/src/components/Radar.tsx` already has a legend, but it's easy to miss: `fontSize={7}` axis labels (`"⅓ ctx"`, `"full"`) and a small panel subtitle `"edge = context full"`.

## Proposal — make the meaning unmissable
- A clearer, larger **legend**: "center = fresh context · **edge = about to compact**" up front, not tiny axis text.
- Mark the **compaction ring** distinctly (a dashed/coloured ring at the threshold) so "the danger line" is something you can watch a blip cross — pairs directly with the radial-scale change in #44.
- A **hover / first-run hint** explaining the encoding (distance = context, size = recent activity, colour = status).
- Consider a one-line caption under the title so the panel self-documents.

## Why it matters
This is a legibility fix, not a new capability — cheap, and it turns a graphic people can't read into one that communicates state at a glance. Companion to #44 (which fixes the underlying radial math); this one makes whatever the math says obvious to a first-time viewer.

SirAllap:
Resolved by #86. The radar rework landed everything this asked for:

- the up-front legend under the chart: **center = fresh context · edge = about to compact**;
- a distinct **dashed compaction-threshold ring** — the danger line a blip actually crosses;
- `⅓ →compact` / `compact` range labels anchored to that ring, so the scale reads as fraction-to-compaction rather than raw model max;
- a hover tooltip that spells out `ctx X / Y (N% · M%→compact)`.

The distance encoding is self-documenting now instead of reading as decorative. Closing as done.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #86, opened 2026-07-20
feat(radar): anchor radius to compaction threshold

### LINKED ISSUE #44: Radar radial scale: anchor to the compaction threshold + non-linear emphasis (enhancement over a0d5cca)
## Status — checked against `git log`

The **base** radar mapping shipped in **`a0d5cca` — _feat(radar): distance from centre = context window used_** (the only scale commit; `Radar.tsx` has just that + the initial release). So "distance = context" already exists.

**This issue is scoped to the enhancement only** — a better radial *scale* (anchor + curve) — which is **not** in the code today. If the enhancement isn't wanted, this is a clean close.

## What's in the code (as of `main`)

`web/src/components/Radar.tsx:41-43`:
```ts
const frac   = a.ctxTokens > 0 ? Math.min(1, a.ctxTokens / a.ctxLimit) : null; // 0..1 of model MAX
const age    = Math.min(1, (now - a.lastSeen) / (5 * 60_000));                  // recency fallback
const radius = 16 + (frac ?? age) * (R - 26);                                   // LINEAR → [16, 78]px
```
Linear, anchored to the model's **raw max** (`ctxLimit` from `ctxLimitOf` in `derive.ts`), edge labelled `"full"`.

## Why the current scale is worth improving

1. **The edge is the wrong anchor.** Compaction fires at a **threshold (~0.85–0.92)**, not the model max. With `frac = used / max`, a session compacts while still ~90% out, so the outer ring — the "about to compact" signal the panel promises — is **never reached** and the warning lands late. Anchor to the threshold so **edge = "at compaction."**
2. **Linear wastes resolution where it matters.** 0→50% looks the same as any idle session; the decisions live in the last stretch. A mild curve that **expands the top end** (e.g. `p ** 0.7`, or piecewise stretching `p > 0.6`) gives the danger zone the radial space it deserves.
3. **Cross-model semantics.** 200k (Claude) vs 1M (Gemini) vs 400k (GPT-5): distance is *fraction-to-compaction*, not absolute size. Arguably correct, but state it explicitly and decide deliberately.
4. **Validate `ctxTokens`.** It's the latest main-turn `input + cache_read + cache_creation` (`derive.ts`) — confirm that matches how each provider counts the live window.

## Proposal
- Add **`COMPACT_FRAC`** (default ~0.9, ideally model-aware) and scale `p = min(1, frac / COMPACT_FRAC)` → edge = compaction threshold.
- Apply a mild easing so the outer band is legible.
- Keep the recency fallback (dimmed) for turn-less sessions.
- Re-label rings to the new thresholds (pairs with #46).

## Acceptance criteria
- A session at the real compaction point sits **on the outer ring**, not ~90% of it.
- 60% vs 95% of the threshold are **clearly** different radii.
- Turn-less sessions still fall back to recency, visibly dimmed.

Self-contained: one file (`Radar.tsx`), pure geometry, no server change. Good candidate for the `derive`/geometry tests in #10.

Solaris-star:
## Summary
Implements the radial-scale enhancement from #44 (and the legibility bits that pair with #46):

- Anchor distance-from-centre to `COMPACT_FRAC` (default `0.9`) so **edge = about to compact**, not raw model max
- Mild `p ** 0.7` easing to expand the outer danger band
- Dashed compaction ring + clearer caption / range labels
- Tooltip shows both raw % of model max and % of the compaction threshold
- Recency fallback for turn-less sessions is unchanged (still dimmed)

Self-contained in `web/src/components/Radar.tsx`.

Closes #44
Related: #46


SirAllap:
Thank you for this! 🙏 It's a faithful, clean take on the compaction-threshold idea, and I really like that you added the `ctxLimit > 0` guard on your own initiative. The dashed danger ring and the tooltip showing both the raw % and the %-to-compact are exactly the kind of legibility I was hoping for.

Two tiny, completely optional observations — please don't feel any obligation, I'm glad to tweak them myself after merge:
- the `compact` range label sits around radius 104 while the dashed ring is at `OUTER` (94), so they're a hair out of alignment;
- there are now two rings fairly close together (94 and 104), which can read a touch busy.

Neither is blocking in the slightest. And whenever you get a sec — no rush — could you confirm `bunx tsc --noEmit` and `vite build` pass in `web/`? We don't have CI on the repo yet (that's on me, not you). Thanks again for picking this up so quickly. 🙌

---

SirAllap:
Heads-up: #89 landed on `main` and touched `Radar.tsx`, which put this branch in conflict — my fault for merging on top of an open PR, so I've resolved it for you rather than making you rebase.

The two changes turned out to be orthogonal and both are kept:
- yours anchors the blip **radius** to the compaction threshold
- `main` tints an **idle** blip by how the session finished

The only overlap was the `return` line. `frac` stays the raw fraction of the model max, since that's what the tooltip and the legend read — only the radius is re-anchored.

Typecheck, tests (235) and the production build are green. Nothing else about your change was altered.

SirAllap:
Made the two legibility tweaks so this doesn't stall: anchored the inner guide rings and both range labels to `OUTER` (the actual blip band) and dropped the redundant solid ring at r=1, so `compact` now sits on the dashed compaction ring and the two-rings-close busy-ness is gone. Confirmed `bunx tsc --noEmit` and `vite build` both pass in `web/`. Thanks for the clean take on this. 🙌

Issue #47, opened 2026-07-19
Docs: an "Extending agentglass / make it yours" guide (point any agent at it, add themes, use the gate as a primitive)


SirAllap:
## The idea

agentglass is a **visibility + control-plane layer** — the thing you point your agents and your harness *at*, not a harness itself. That distinction matters: you don't replace your existing setup, you feed it in. But it isn't spelled out anywhere, and the extension surfaces that **already exist in the code** aren't collected in one place, so adapting agentglass looks like a fork-and-hack job when it usually isn't.

## What already exists (worth documenting, not building)

- **Feed any agent, no Claude required** — `POST /ingest` (hook payloads) and the **OTLP receivers** `POST /v1/traces` + `/v1/logs` (`server/src/otlp.ts`, `otlp_pb.ts`) map any OpenTelemetry `gen_ai.*` exporter into events. This is the "point *anything* at it" story (Codex, Gemini CLI, Bedrock, LangChain, LiteLLM…) and it's underused.
- **The Claude Code integration** — `hooks/` (`install_hooks.py`, `send_event.py`, `gate_event.py`): what each hook sends and how to wire a project.
- **The gate as a generic approval primitive** — `POST /gate` (`server/src/gate.ts`) is just "hold this action until a human decides." Anything, not only Claude, can long-poll it — a reusable building block for someone's own harness.
- **Config & scoping** — `~/.config/agentglass/config.json`, `AGENTGLASS_ROOT` / `repoDirs`, the write-gates / token (`server/src/config.ts`, `auth.ts`).
- **Add a theme** — the shape in `web/src/lib/themes.ts` (copy one, edit values).
- **How "live" works** — the WS event stream + buffered flush (`web/src/lib/useLive.ts`, `getRecent` initial frame), so the "does it update live?" question has a documented answer.

## Proposal

A single **`docs/EXTENDING.md`** (linked from the README):

1. **"Point any agent at agentglass"** — the OTLP + `/ingest` contract, with a minimal working example for a non-Claude exporter.
2. **"Use the gate in your own harness"** — the `/gate` request/response shape and the fail-open/closed contract.
3. **"Make it yours"** — add a theme, scope to your projects, the config surface.
4. **"How it stays live"** — a short architecture note (scanner + hooks/OTLP → SQLite → WS).
5. A one-paragraph **"dashboard vs harness"** framing up top.

## Why it matters
[...]

PR #87, opened 2026-07-20
docs: Extending agentglass / make it yours guide

Solaris-star:
## Summary
Docs-only guide requested in #47: a single `docs/EXTENDING.md` that surfaces extension points already in the tree (no new code).

- Point any agent at agentglass (`POST /ingest`, OTLP `/v1/traces` + `/v1/logs`)
- Use the gate as a generic approval primitive
- Themes, config/scoping, auth/write gates
- How the live WS path stays live (`useLive` buffering)
- Dashboard vs harness framing

Linked from README contents + Contributing.

## Test plan
- [x] Docs-only; no runtime change
- [ ] Spot-check curl examples against a local `bun run dev` if desired

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ gate. No install, no server. *(Everything there is fake; it's a showcase.)*
 - [Control plane — approve / deny remotely](#control-plane--approve--deny-tool-calls-remotely-opt-in)
 - [Any provider — via OpenTelemetry](#any-provider--via-opentelemetry-openai-gemini-bedrock-)
 - [Configuration](#configuration-env) · [API](#api) · [Architecture](#architecture)
+- [Extending / make it yours](docs/EXTENDING.md)
 - [Contributing](#contributing) · [License](#license)
 
 ---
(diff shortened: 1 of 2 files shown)


SirAllap:
Thank you for writing this — it's genuinely well done. 🙏 I went through the technical claims against the code and the `/ingest` fields, the gate timeout bounds, `AGENTGLASS_CHAT_DISABLED`, and the whole "how it stays live" section all check out. Lovely, careful work.

One small thing, and please take it or leave it — no pressure at all: the line *"Intake routes (`/ingest`, OTLP) honour the token when set"* is actually the opposite of how it works now — those routes are intentionally **tokenless** (local hooks and OTel exporters have no way to carry a secret), while everything else needs the token. Maybe something like *"stay tokenless on purpose; everything else needs the token when set"* would match the current behaviour.

Truly happy to fix that one line myself before merging if you'd rather not bother — just say the word. Thanks so much for putting this together. 🙌

---

SirAllap:
Applied your one-line fix and merging so this doesn't stall — the intake routes line now reads that `/ingest` and OTLP stay tokenless on purpose (local hooks and OTel exporters can't carry a secret) while everything else needs the token when set. Reviewed the rest against the code and it holds up; genuinely nice guide. 🙌

Issue #83, opened 2026-07-19
Docs: the README describes an older app (timeline, tool output, resume, images, skills, scope-as-boundary all missing)


SirAllap:
A run of changes has left the README describing an app that no longer matches the one you install. The gap is now big enough that someone arriving at the repo forms the wrong idea of what it does.

## Behaviour the README doesn't mention

- **Session timeline** (#57) — the conversation interleaves tool runs with messages, so a session shows what the agent *did*, not only what it said.
- **Tool output** (#73) — each run carries the head of its result, so a failing test is distinguishable from a passing one without leaving the panel.
- **Subagent drill-down** (#63) — a fleet of subagents reports the parent's session id; clicking one filters the thread to it.
- **Resume** (#50, #61, #64) — a finished session can be picked back up from the fleet view or the Chats panel, with its transcript replayed.
- **Image attachments** (#68, #77) — images go to the model as image blocks; other files are quoted into the message.
- **Skills via `/`** (#78) — the composer lists them; slash commands were always enabled in `-p`, they just weren't discoverable.
- **Scope as a boundary** (#67, #72) — this one matters most, because it's a *behaviour change*: with a project open, git writes, the terminal and chat are refused outside it. The README has a line about it now, but the "work across repos by scoping to the parent folder" answer deserves to be visible rather than buried in an error string.
- **`AGENTGLASS_FS_BROWSE_DISABLED`** — documented in the env table, but the security model section doesn't mention that directory completion exists as a surface at all.

## Screenshots

Every screenshot predates the header rework (#52) — they show the old app-filter layout and no project control. The diff shots also predate syntax highlighting working in the desktop app (#76), so they show monochrome code, which now misrepresents the product in the least flattering direction possible.

Regenerating them is its own small task: the demo mode (`VITE_DEMO=1`, `web/src/lib/demo.ts`) is a complete fabricated dataset and is the right source, so no scratch repos or dummy containers are needed. Worth knowing before anyone rebuilds that scaffolding by hand.

## Why it's worth doing properly

The README is what the stars are looking at. Right now it undersells the app and, on scoping, actively misdescribes it.

PR #88, opened 2026-07-20
docs: refresh README for timeline, tool output, resume, images, skills

Solaris-star:
## Summary
Updates the README to describe features that shipped after the last docs pass (#83):

- Session **timeline** interleaving tool runs with messages
- **Tool output** heads on tool cards (failing vs passing visible in-panel)
- **Subagent drill-down** via parent session id
- **Image attachments** sent as image blocks; other files quoted
- **Skills via `/`** in the composer
- Clarify **scope as a boundary** (terminal/git writes refused outside root)

Docs-only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,15 @@ in a terminal — with its full context intact. Sessions that are still running
 are listed but can't be picked: a claude session has a single owner, and a
 second writer on the same transcript corrupts its history.
 
+**What a session shows:** the conversation is a **timeline**, not only a chat
+log. Tool runs interleave with messages, each tool card carries the head of its
+output (so a failing test is distinguishable from a passing one without leaving
+the panel), and **subagents** report the parent's session id — click one to
+filter the thread to it. Images are sent to the model as image blocks; other
+files are quoted into the message. Type `/` in the composer to list and insert
+skills/commands (slash commands are enabled in `-p`, they just weren't
+discoverable).
+
 ---
 
 ## Why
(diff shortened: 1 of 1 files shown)


SirAllap:
Thanks for keeping the README current — this is really helpful, and the session/timeline additions read nicely. 🙏

One gentle note, no obligation whatsoever: the reworded scope-boundary line ties the outside-root refusal to the `AGENTGLASS_GIT_WRITE_DISABLED`/`AGENTGLASS_TERMINAL_DISABLED` knobs, but the scope boundary actually refuses writes outside the root on its own whenever a project is open — those knobs are a separate global off-switch. The original phrasing captured that distinction a bit better, so it might be worth keeping it so the "boundary, not cosmetic" point stays strong.

And whenever you have a moment — no rush — could you double-check the "What a session shows" features (tool-output heads, subagent drill-down, `/` skills) are all in `main` exactly as described? Just want the docs to line up precisely.

Either way, thank you for this — and I'm very happy to make these little tweaks myself if you'd prefer. 🙌

---

SirAllap:
Applied the tweak myself so this doesn't stall — restored the scope-boundary phrasing so the outside-root refusal reads as intrinsic to the boundary (it happens on its own whenever a project is open), with the `AGENTGLASS_GIT_WRITE_DISABLED`/`AGENTGLASS_TERMINAL_DISABLED` knobs called out as a separate global off-switch. Also re-checked the three "What a session shows" claims against `main`: tool-output heads (`ToolRow`), subagent drill-down via parent `session_id` (`derive.ts`), and `/` skills in the composer (`ChatPanel`) are all present as described. Thanks for keeping the README current — merging. 🙌

PR #93, opened 2026-07-20
fix(desktop): macOS install and window controls

myst3k:
## What does this PR do?

Makes the desktop app usable on macOS: `make desktop` now bundles an `.app` and `make desktop-install` installs it to `~/Applications` — the installer previously hardcoded the Linux x86_64 sidecar name, so it always failed with "run 'make desktop' first" on a Mac. The frameless window gains native traffic lights (`titleBarStyle: Overlay`) and can be moved by dragging its header: drag regions in the header plus the `core:window:allow-start-dragging` permission, which Tauri's defaults omit — so this also enables header-dragging on Linux, where WM modifier-drag had been masking its absence. The Linux install flow is otherwise unchanged, now triple-agnostic (works on ARM, no rustc needed at install time).

## How was it tested?

- `bunx tsc --noEmit` in `web/` and `server/`, `bun test` in both (170 + 65 green), `bunx vite build`, `bun run build:demo`, and `bun scripts/smoke.ts` (headless Chromium boots the production bundle, no console errors)
- Built and installed on an Apple Silicon Mac: launches from Spotlight and the CLI symlink, the bundled server starts, traffic lights work, the window drags by its header and resizes as before
- Linux install path regression-tested by running `install-local.sh` in a sandbox (stubbed `uname`, throwaway `$HOME`): file layout identical to the current release — binary + sidecar under `~/.local/share/agentglass`, `.desktop` launcher, icons
- Release CI unaffected: `desktop-binaries.yml` passes `--bundles dmg` explicitly, which overrides the new macOS config's `app` target

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ src-tauri/bin/
 src-tauri/icon-*.png
 src-tauri/gen/schemas/
 .claude/
+*.bun-build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,8 +30,8 @@ python3 hooks/seed_demo.py   # populate with demo data
 - **`hooks/`** — stdlib-only Python; keep it dependency-free.
 - **`src-tauri/`** — the Tauri v2 desktop shell (Rust). `make desktop` compiles
   the Bun server to a standalone sidecar, builds the web bundle, and runs
-  `tauri build`; `make desktop-dev` runs against the live dev server. Linux only
-  for now (`.deb`).
+  `tauri build`; `make desktop-dev` runs against the live dev server. Linux
+  (`.deb`) and macOS (`.app`).
 
 ## Ground rules
 
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ desktop: desktop-server desktop-web ## Build the desktop app (icon + native wind
 desktop-dev: desktop-server ## Run the desktop app against the live dev server
 	bunx tauri dev
 
-desktop-install: ## Install the built app for this user (~/.local, no root)
+desktop-install: ## Install the built app for this user (no root)
 	src-tauri/install-local.sh
 
 # Open the cockpit for ONE project: only that repo (and its worktrees) appear,
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ the server goes with it.
 
 ```bash
 make desktop            # build: compiles the server sidecar + web, then `tauri build`
-make desktop-install    # install for this user under ~/.local (no root)
+make desktop-install    # install for this user (no root)
 ```
 
 Then launch **agentglass** from your desktop menu, or `agentglass` from a shell.
@@ -279,9 +279,7 @@ directory instead:
 make desktop-open DIR=~/code/my-project   # or: agentglass ~/code/my-project
 ```
 
-> Built for **Linux** today (`.deb` bundle, `x86_64`). The `src-tauri/` Tauri v2
-> shell + Bun sidecar is portable in principle, but only the Linux build is
-> wired up for now.
+> Works on **Linux** and **macOS**.
 
 ---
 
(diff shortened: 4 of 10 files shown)


SirAllap:
Thank you for this — and for building it on a platform I don't own and can't test. macOS support has been an open question since the desktop app landed and nobody had actually tried it until now, so you found this the hard way.

The diagnosis is the part I appreciate most. The hardcoded `x86_64-unknown-linux-gnu` sidecar name failing with "run `make desktop` first" is a genuinely misleading error — it tells you to do the thing you just did. And `current_exe` returning the CLI symlink on macOS rather than the resolved path is the kind of detail that costs an evening to find and one line to fix.

The `data-tauri-drag-region` note in your commit message — one per container, because the attribute only applies to the exact element clicked — saved me from "wait, why four of these?" And you were right that this fixes Linux too; WM modifier-drag had been quietly covering for the missing permission the whole time, so I never noticed the header wasn't draggable on its own.

You also checked that `desktop-binaries.yml` passes `--bundles dmg` explicitly, so the new `targets: ["app"]` only affects local builds. I verified the same before merging. Getting the release path right when your change doesn't touch it is easy to skip.

I pulled the branch and ran it on Linux. It sits directly on current `main`, so this is your commit as-is, no merge needed:

- `bun test` — 178 pass in `server/`, 65 pass in `web/`, 0 fail (server is up from your 170; `main` gained tests after you branched)
- `bunx tsc --noEmit` clean in both packages, `bun run build` clean
- `install-local.sh` with a stubbed binary and a throwaway `$HOME`: layout identical to the current release — `agentglass` + `agentglass-server` under `~/.local/share/agentglass`, symlink in `~/.local/bin`, `.desktop` launcher, icons at 32/64/128

**Two small things I'm going to handle myself in a follow-up** — mentioning them so you see them land, not asking you to do anything:

The sidecar glob picks the first match, and `src-tauri/bin/` is gitignored so it accumulates across builds. I reproduced it on my x86_64 box by dropping a stale aarch64 sidecar next to the real one:

```
$ find src-tauri/bin -name 'agentglass-server-*-linux-*' -type f | head -n1
src-tauri/bin/agentglass-server-aarch64-unknown-linux-gnu
```

Wrong arch, installed silently, fails at exec time. Yours is strictly better than the hardcoded name — I just want to filter on `uname -m` so that can't happen. Same line, `find | head -n1` under `set -euo pipefail` can also abort with no output if `find` takes SIGPIPE. Both need two triples in `bin/` to trigger, so neither is something you'd hit on a clean machine.

**One thing I wanted to ask, and it's more question than observation.** I saw you tested the CLI symlink. Did you run it as `agentglass` from the shell, or by invoking the path directly? I ask because `~/.local/bin` isn't on the default macOS PATH — `/etc/paths` only carries absolute system paths, so a directory inside your home can't get there unless something edited your shell rc (uv and pipx both do). If it works on your machine, you already have it set up from before. If you tested it by direct path, then the script is printing `command  agentglass` for people who won't be able to run it, and I'll fix that with a PATH check in the output. Either way there's nothing to change on your end — I just want to confirm before touching behavior you verified.

`make desktop-open` works regardless, since you symlinked the absolute path under `~/.local/share` too. Nice catch.

This is going in. Branch protection wants an approving review before it can merge, which is on my side to sort out, not yours. This is the first time the desktop app works on macOS, and it works because you had the machine and took the time. Thank you.

SirAllap (approved):
Verified on Linux, and the macOS side is verified by the author on hardware I don't have. Details in the review comment above.

myst3k:
Confirmed your theory: I ran it as `agentglass` by name, and it only resolved because `~/.local/bin` was already on my PATH — my own doing, from a previous tool's post-install instructions (Claude Code's installer has you add it manually; my shell history shows me pasting exactly that line). Stock macOS wouldn't have it, so a PATH check in the install output sounds right. Both sidecar-glob fixes make sense too — thanks for catching what a clean machine couldn't show me.

PR #112, opened 2026-07-20
fix(ui): clarify tool latency measurement

### LINKED ISSUE #110: Label what tool latency (p50/p95) actually measures
## The ambiguity

The tool-latency percentiles are computed by pairing `PreToolUse` → `PostToolUse` timestamps (`server/src/db.ts`), so they measure **wall-clock tool duration** — how long the tool actually ran on the machine, end to end. That is the right thing to measure, but the UI just labels it "tool latency p50/p95" with no qualifier, so a reader can reasonably assume it means the **LLM API round-trip**. Those are very different numbers, and someone judging the tool against the wrong baseline will draw wrong conclusions.

## Proposal

- **Label it explicitly** in the stats panel: e.g. "tool wall-clock (p50/p95)" plus a tooltip: *"time from PreToolUse to PostToolUse — how long the tool took end to end, not the model round-trip."*
- Optional follow-up: if turn/inference latency is available from the transcript or OTLP spans, expose **model round-trip latency** as a separate metric so both questions have an answer.

## Why it matters
The latency numbers are only trustworthy if it is clear what they clock. This is a small labeling fix that stops a confident misread.

## Entry points
The tool-latency section of the stats UI (`web/src/components/…` where `ToolLatencyStat` renders), and `server/src/db.ts` where `duration_ms` is derived from the Pre/Post pairing.

mgalore:
## What does this PR do?

Fixes #110. Clarifies that the performance panel’s p50/p95 values measure end-to-end tool wall-clock duration from `PreToolUse` through `PostToolUse`, rather than model/API round-trip latency.

## How was it tested?

- `bun test` (267 passing)
- `make typecheck`
- `bun run build`
- `git diff --check`

### DIFF
--- a/web/src/components/Latency.tsx
+++ b/web/src/components/Latency.tsx
@@ -9,8 +9,11 @@ export const Latency = memo(function Latency({ stats }: { stats: StatsSummary | 
   const max = Math.max(1, ...tools.map((t) => t.p95_ms));
 
   return (
-    <Panel eyebrow="Performance" title="Which tools are slow" right={<span className="text-[10px] t-dim2">p50 · p95</span>}>
+    <Panel eyebrow="Performance" title="Tool wall-clock duration" right={<span className="text-[10px] t-dim2">p50 · p95</span>}>
       <div className="overflow-auto h-full pr-1">
+        <p className="t-dim2 text-[10.5px] leading-snug mb-2">
+          End-to-end tool duration from PreToolUse to PostToolUse, not model/API round-trip latency.
+        </p>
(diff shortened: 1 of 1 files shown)


SirAllap:
Merged, thank you! 🎉

This is exactly what #110 was after, and the wording is spot on: that is precisely how `duration_ms` gets paired, from `PreToolUse` through `PostToolUse`. Clean, well scoped and nicely tested, and fast too. Really appreciate you jumping on it.

Welcome aboard, and thanks for the contribution. 🙌

---

PR #115, opened 2026-07-21
feat(server): serve the built web UI from the API port (single-port deploys)

icey2488:
## What does this PR do?

Teaches the server to serve the built web UI (`web/dist`) from the API port when a build exists — so a headless deploy can drop the second process entirely: one `bun` process, one port, dashboard and API on the same origin. The desktop shell already proved the server side of this arrangement (it hosts `web/dist` over loopback HTTP itself); this brings the same thing to server-only installs, e.g. a systemd unit on a home box.

Mechanics, kept deliberately narrow:

- **Exact files only, resolved traversal-safe** under `web/dist` (same clamp idiom as `safeAbs` in `git.ts`), MIME from the extension, far-future cache on the content-hashed `assets/`, `no-cache` on everything else. API and WS routes never collide — they map to no real file, so they fall through untouched.
- **SPA fallback runs last**, after every API route has declined, and only for a `GET` that asks for html — `curl` on a bad path still gets the JSON 404.
- Static files sit ahead of the token gate (the bundle is the same public code that ships in this repo, and the `?token=` flow needs `index.html` + assets to load before the app can pick the token up); every data route stays gated. Bind semantics are untouched.
- `index.html` goes out with a small planted marker, and `api.ts` resolves its API base **same-origin** when it sees one. Serve-time rather than build-time, so the identical bundle keeps resolving `:4000` under vite dev/preview and the desktop shell's static server. `VITE_CW_SERVER` still overrides everything.
- No `web/dist` (plain `bun run dev`, hooks-only installs) → the server is API-only, exactly as before.

## How was it tested?

- `bunx tsc --noEmit` clean in `server/` and `web/`; `bun test` green in both (190 server / 84 web, including new `server/test/webui.test.ts` covering the traversal clamp — encoded `..`, NUL bytes, malformed escapes — and the marker injection).
- Live on the WSL install this stack runs on daily: `/` serves the marker-carrying `index.html`, hashed assets come back `text/javascript` + `immutable`, `favicon.svg` as `image/svg+xml`, `/stats` and `/health` unchanged, a deep-link with `Accept: text/html` gets the SPA, `curl` on an unknown path gets the JSON 404, and `--path-as-is /assets/%2e%2e/%2e%2e/package.json` is refused.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,6 +210,18 @@ Open **http://localhost:6180**. Then see it move without wiring a real project:
 python3 hooks/seed_demo.py            # streams demo agents for ~30s
 ```
 
+**Single-port deploys** (a headless box, a systemd unit): build the UI once and
+the server serves it itself — one process, one port, API and dashboard on the
+same origin:
+
+```bash
+bun run build                         # web/dist
+cd server && bun run src/index.ts     # dashboard AND API on :4000
+```
+
+When `web/dist` doesn't exist (plain `bun run dev`), nothing changes — the
+server is API-only and the vite dev server owns the UI as before.
+
 Prefer `make`? Every entry point is a described Makefile target — `make help`
 lists them all (`make dev`, `make setup`, `make demo-feed`, `make desktop`, …),
 and the in-app terminal (`t` → **⚙ commands**) surfaces the same list, ready to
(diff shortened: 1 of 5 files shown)


SirAllap:
Thank you for this — it's a genuinely thoughtful contribution. Single-port deploys are something people have wanted for the headless/systemd case, and you built it exactly the way it should be built: the traversal clamp mirrors the `git.ts` threat model (with tests to match), the gate ordering is reasoned out rather than assumed, and the serve-time marker keeps the same bundle working under vite and the desktop shell. The care in the comments made it easy to review.

Merging. Really appreciate the work. 🙏

---

PR #137, opened 2026-07-21
fix(desktop): stop the app instead of shooting every process it owns

### LINKED ISSUE #132: install-local.sh mass-pkills the app, wedging it instead of stopping it
## What happens

Rebuilding over a running install leaves the old window on screen, frozen, for a long time. Sometimes the process never dies at all: after one rebuild in this repo, the previous instance was still alive 34 minutes later, holding deleted inodes, and needed a manual `kill -9`.

## Why

`electron/install-local.sh`:

```bash
if pgrep -f "$APP/agentglass" >/dev/null 2>&1; then
  pkill -f "$APP/agentglass" 2>/dev/null || true
  sleep 2
  pkill -9 -f "$APP/resources/agentglass-server" 2>/dev/null || true
  sleep 1
fi
```

Two problems.

**The pattern is too broad.** `$APP/agentglass` appears in the cmdline of every Electron process: main, `--type=gpu-process`, `--type=zygote`, `--type=renderer`. So all of them get SIGTERM simultaneously, and the main process is trying to run its own quit path (`stopSidecar()` then `app.quit()`) while its children are being torn out from under it. That is the wedge, and the frozen window is what it looks like.

Measured against the main process alone, the app is well behaved:

```
t+1s: still alive
t+2s: gone       (no survivors, sidecar port released)
```

**The wait is a guess, not a check.** `sleep 2` then proceed. If the app is still up, the script starts `rm -rf`-ing and copying over a live install, which is exactly the "app running on deleted inodes" state the comment above it warns about.

## Expected

1. Signal the main process only, with a pattern anchored to the exact cmdline.
2. Poll until it is actually gone, with a bounded wait rather than a fixed sleep.
3. Escalate to `-9` for whatever is left, including strays and the sidecar, and only then touch files.

Found while rebuilding after #127 and #130.

SirAllap:
## What does this PR do?

`install-local.sh` stopped a running install with `pkill -f "$APP/agentglass"`. That path is in the command line of **every** Electron process, so main, zygote, gpu-process, utility and renderer all took SIGTERM at once, and the main process was left running its quit path (`stopSidecar()` then `app.quit()`) with its children pulled out from under it. That is the wedge: a frozen window that outlives the rebuild, once still alive 34 minutes later on deleted inodes with the sidecar still holding the port. The `sleep 2` afterwards was a guess, and the script would `rm -rf` and copy over whatever was still running.

Process handling moves into `electron/appctl.sh`:

- pids come from `/proc/<pid>/exe`, not the command line: that resolves an instance started through the `~/.local/bin` symlink, tells main from helper (helpers all carry `--type=`), and still recognises an install whose files were already replaced, which reads as `(deleted)`
- SIGTERM goes to the **main process alone** and Electron shuts itself down
- the wait polls for an empty install instead of sleeping, escalates to SIGKILL on anything left after 10s, and returns non-zero if something survives, at which point `install-local.sh` refuses to install rather than copying over a live app
- candidates are narrowed with one `grep` over `/proc/*/cmdline` before any `readlink`, because a readlink per pid cost the best part of a second per call and turned the polling loop into the stall it was meant to remove

Closes #132

## How was it tested?

- New `server/test/install-stop.test.ts`: builds a fake install (a real copied binary in a real directory) and runs real processes out of it, because the bug was entirely about which processes a pattern matches. Covers the four helper shapes Electron actually spawns, a second install that must be left alone, a healthy shutdown, a process that ignores SIGTERM (must be escalated, not slept through), an orphaned sidecar, and the nothing-running no-op.
- `bun test` (server + web): **377 pass, 0 fail**.
- Against the **real app** on this machine: launched the installed build, confirmed 8 processes with exactly one main, then `stop_app` sent SIGTERM to that one pid and every process was gone in **0.2s** with no escalation and no survivors.
- Full `install-local.sh` run end to end afterwards: packaged, installed, and the app launches from the new install.

### DIFF
--- a/electron/appctl.sh
+++ b/electron/appctl.sh
@@ -1,0 +1,116 @@
+#!/usr/bin/env bash
+# Finding and stopping an installed agentglass desktop instance.
+#
+# Sourced by install-local.sh (and exercised by server/test/install-stop.test.ts),
+# which is why it is a library rather than inline: the interesting part is not
+# the install, it is knowing which processes belong to this install and waiting
+# for them to actually be gone.
+#
+# Expects APP to be set to the install directory (…/agentglass-desktop).
+
+# The resolved binary behind a pid, or nothing.
+#
+# /proc/<pid>/exe rather than the command line, for three reasons this script has
+# been bitten by: an instance launched through the ~/.local/bin symlink has a
+# different argv[0]; every Electron helper repeats the same path in its own argv,
+# so a cmdline match cannot tell main from renderer; and an instance whose files
+# were already replaced still answers here, with a "(deleted)" suffix, which is
+# precisely the state we are trying to avoid creating.
+_exe_of() {
+  local link
+  link=$(readlink "/proc/$1/exe" 2>/dev/null) || return 1
+  printf '%s\n' "${link% (deleted)}"
+}
+
+# Anything whose command line mentions us at all — the cheap first pass.
+#
+# One grep over /proc beats a readlink per pid by a wide margin, and it matters:
+# app_pids is called in a polling loop, and walking every process on a busy
+# machine took the best part of a second per call, which turned a bounded wait
+# into a visible stall. This over-matches on purpose (a shell sitting in a
+# directory named agentglass lands here too); app_pids does the precise part.
+_candidate_pids() {
+  local f
+  for f in $(grep -la -e agentglass /proc/[0-9]*/cmdline 2>/dev/null); do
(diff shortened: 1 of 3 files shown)


SirAllap:
Superseded: #136 landed the same fix while this was in CI. Reopened as the delta on top of it, since matching by command-line shape leaves a gap this one closes (a main process started with a flag matches nothing, so the stop block is skipped entirely and the script installs over a live app).

PR #175, opened 2026-07-22
perf(server): stop the one thread freezing the app

SirAllap:
## What does this PR do?

The terminal froze — mid-keystroke, for up to 2.8s at a time, several times a minute. This is the whole hunt, and every commit in it is one measured cause.

**The root of it**: this server has one thread, and git and docker were being spawned synchronously on it. A `spawnSync` in a poll handler is a freeze with a name. Everything expensive is async now, behind a pool that caps how many subprocesses can be in flight at once, with a timeout so a git that never answers releases its slot instead of holding it for good.

**The offenders, in the order they were caught**: the repo sweep storming the machine every five seconds; the filter dropdown, which was the single worst freeze in the app; the scope filter scaling with the number of worktrees; the update check dialling GitHub on the thread that carries the terminal; the skills scan; four expensive git reads; a sync fan-out costing two subprocesses per worktree; and two panels racing to re-derive the same dirty counts.

**The thing that made it findable**: `loopwatch` watches the event loop and names what blocked it. Before that, "the terminal feels laggy" was a feeling. It also learned to blame the work that actually blocked rather than the poll that happened to arrive while it was blocked, which is a different and much less useful answer.

**The terminal wins**: background work stands back while you are typing.

Also here, because its own test needs a fast sweep to pass: the squash sweep never finished, so it never found anything — it yielded to a timer that a later poll then cancelled.

## How was it tested?

- `scripts/perfbudget.ts` fails the build when the loop would block long enough to be felt. It is wired into CI, so this class of bug cannot come back quietly.
- `scripts/soak.ts` answers "does it grow?" with a measurement rather than an opinion.
- New `server/test/spawnpool.test.ts` over the cap, the queueing and the release, and `server/test/git-timeout.test.ts` over a hung git releasing its slot.
- `server/test/loopwatch.test.ts` covers the attribution.
- `cd server && bun test`: 375 pass. `cd web && bun test`: 178 pass. Both typechecks clean, `vite build` clean.

Third of six, on top of #2. The branches above depend on `gitAsync` and the pool landing here first.


SirAllap:
Closed automatically when its base branch was deleted by the merge of #174, before GitHub could retarget it. Reopened as a new pull request against `main` — same branch, same commits.

PR #237, opened 2026-07-23
feat(control): drive the dashboard UI from an external controller (POST /control)

Yoshiofthewire:
## What

Adds **`POST /control`** — a one-way channel that lets anything on the machine (a Stream Deck, a phone, a script) drive the dashboard's own UI: switch workspace views, toggle the workspace, open panels, cycle the theme, and zoom — without touching the keyboard.

## Why

The dashboard is fully keyboard-navigable, but that navigation is **client-side React state with no external trigger** (`web/src/App.tsx`'s keydown handler). Everything else an external controller would want — approve/deny (`/gate/decide`), monitoring (`/stats`, `/insights`, `/gate/pending`, `/stream`) — already exists. UI navigation was the one gap. This is the bridge that fills it, and it's the groundwork for the phone-friendly remote in #7.

## How

- The server validates a `ControlCmd` against a **closed set** (`server/src/control.ts`) and rebroadcasts it as a `control` frame on the `/stream` socket every tab already holds.
- `useLive` hands the frame to `App` via a small `controlBus` (same pub/sub shape as `gitBus`), which runs it through **the very setters the keyboard handler already uses** — so external and keyboard navigation are one code path, not two.
- `ViewId` moves to `shared/types.ts` so the server and UI validate against one set (`views.ts` re-exports it).

## Safety

It grants **no capability the keyboard doesn't**: a command changes only *what is shown*, never the fleet. So it needs no gate beyond the `localOrigin` + token checks the whole surface already carries — a no-`Origin` local caller is admitted on a loopback bind (like `hooks/gate_event.py`), and it requires `Authorization: Bearer <token>` when `AGENTGLASS_TOKEN` is set. A malformed or unknown command is rejected with `400` before anything is broadcast.

## `ControlCmd`

| `cmd` | Fields | Effect |
|---|---|---|
| `view` | `to`: git\|diff\|pr\|docker\|term\|chat | open the workspace on that view |
| `workspace` | `open?`: bool | toggle / set the workspace overlay |
| `esc` | — | close panels / workspace |
| `open` | `what`: stats\|skills\|search\|help\|palette | open that panel |
| `theme` | `name?` or `dir?`: 1\|-1 | pin a palette / step the list |
| `zoom` | `dir`: 1\|-1\|0 | zoom in / out / reset |

## Tests

- New `server/test/control.test.ts` (14 cases) pins the validator's accepted sets and rejections.
- Full suites green: **server 439 pass**, **web 217 pass**; `tsc --noEmit` clean in both workspaces.
- Manual e2e: a valid `POST /control` broadcasts a `control` frame over `/stream`; unknown command / bad view id → `400`; a foreign `Origin` → `403`.

## Docs

New section 3 in `docs/EXTENDING.md` and a row in the README API table.

This is stage 1 of a Stream Deck + integration; the hardware daemon will live in its own repo and needs only this endpoint plus the existing `/gate` and `/stats`.


SirAllap:
Thank you for this — and welcome, it's a genuinely thoughtful first contribution. A UI-navigation bridge is something the Stream Deck idea in #7 has been waiting on, and you built it exactly the way it should be built: `controlBus` mirrors the `gitBus` pub/sub already in the tree, the command runs through the very setters the keyboard handler owns so there's one navigation path instead of two that drift, and the validator is treated as a real trust boundary — a closed set, unknown input turned away with a `400` before anything is broadcast, and `control.test.ts` pinning it. Moving `ViewId` into `shared/types.ts` so the server and UI name one set is the right instinct, and the care in the comments made it easy to review.

I approved the CI run — first-time contributions need that manually — and it's green across the board. Merging. Really appreciate the work, and I'd love to see the hardware daemon when you get to it. 🙏

---

PR #243, opened 2026-07-24
fix: recognize Windows-style paths in scope check, path completion, and env-var splitting (fixes #188)

### LINKED ISSUE #188: Windows: path handling is POSIX-only in four places, and scoping a project breaks everything inside it
## What happens

Four places assume POSIX path syntax. On Windows each one fails silently or confusingly, and none of them is more than a one-line fix. They are filed together because fixing one without the others still leaves the app unusable.

**1. `PATH`-style env vars split on a literal `":"`**

`server/src/config.ts:217` and `server/src/gitwork.ts:534`

```ts
(process.env.AGENTGLASS_REPO_DIRS || "").split(":").filter(Boolean)
```

`C:\Users\x\code` becomes `["C", "\Users\x\code"]`. Both `AGENTGLASS_REPO_DIRS` and `AGENTGLASS_REPOS` are unusable on Windows.

`server/src/transcripts.ts:41` already does this correctly with `path.delimiter` for `AGENTGLASS_PROJECTS_DIR`, with a comment explaining why. Same fix, unapplied in two places.

**2. Scope prefix tests hardcode `/`**

`server/src/config.ts:117-118` (`inScope`), `server/src/db.ts:374`, `server/src/docker.ts:199`

```ts
if (p === scope || p.startsWith(scope + "/")) return true;
```

`resolve()` returns backslashes on Windows, so the scope root matches itself but **every path inside it fails**. Open a project and git writes, terminal and chat are all refused inside the user's own repo, with an error telling them to open the parent folder, which does not help. This is the worst of the four.

**3. Path completion rejects drive letters**

`server/src/fsbrowse.ts:77` (and `:79`), mirrored client-side at `web/src/components/ProjectPicker.tsx:52`

```ts
if (!raw.startsWith("/") && !raw.startsWith("~")) return null;
```

A drive letter passes neither guard, so "…or type a folder" autocompletes nothing on Windows. Typing the full path and pressing Enter still works, since `POST /workspace` goes through `resolve()` + `statSync()`. Discoverability, not function.

**4. `TOO_BROAD` is a POSIX-only list**

`server/src/gitwork.ts:309`

```ts
const TOO_BROAD = new Set(["/", "/home", "/mnt", "/media", "/run", "/usr", "/opt", "/var", "/tmp", "/etc", "/srv"]);
```

The guard that stops `codeRootsOf` from adopting a whole disk as a scan root knows no Windows roots. `C:\` and `C:\Users` sail through, so a repo at `C:\Users\x\proj` contributes `C:\Users` as a code root and the walk covers every user profile on the box. Inert today only because roots are empty on a fresh install (#194), which means it starts biting exactly when the user fixes that.

## Expected

1. `path.delimiter` in both env splits.
2. Compare with `path.sep`, or normalise both sides to `/` before the prefix test. Whichever, the three call sites should share one helper rather than repeat the expression a fourth time.
3. Accept `/^[A-Za-z]:[\\/]/` in both guards, server and client.
4. Add the Windows roots, or better, decide "too broad" structurally (depth from the drive/mount root) instead of by name.

## Notes

From a Windows portability audit against `55f7b20` (v0.5.0), read plus an observed NSIS install on Windows 11. All four line references verified on `main`.

Worth a test each: they are exactly the kind of thing that regresses invisibly on a Linux dev box.

SpiliosDimakopoulos:


SirAllap:
Welcome, and thank you, this is a genuinely strong first contribution. #188 is a fiddly one: it is four separate POSIX assumptions filed together, and the failure only shows up on a machine none of us has in front of us. You handled that well.

The things that made it a pleasure to review:
- Making `sep` injectable on `isWithin` so the Windows behaviour is exercised directly from a POSIX CI run, rather than needing an actual Windows box. That is exactly the right move for a bug like this, and the test file pins both branches.
- You matched the house comment style so closely I had to check which lines were yours.
- The fixes themselves are correct: `path.delimiter` on both env splits, the drive-letter guard mirrored server and client, and `inScope` routed through the new helper.

One small thing, and I mean small, so please read it as optional. The issue asked for the scope prefix test to be shared by all its call sites, and you set that up perfectly by exporting `isWithin`. There is just one more site that still hardcodes it: `db.ts:422` (`...p.startsWith(r + "/")`). Those are `resolve()`-derived host paths, so it is the same Windows bug you already fixed in `config.ts`, one layer over. Wiring `isWithin` in there (plus a line in the test) would fully close #188.

Two notes so this does not feel like homework:
- `docker.ts` has the same shape but I would leave it: those are container working directories, always Linux paths, so the hardcoded `/` is correct there.
- Point 4 in the issue (`TOO_BROAD`) is already moot, that code was removed in a refactor after the issue was filed, so there is nothing to do for it.

And one bit of pure housekeeping, again only if you feel like it: the repo uses a short What / How template for PR descriptions (there is a small bot that nudges for it). A couple of lines on what the change does and how you checked it makes the intent easy to read straight from the history later, and helps the next person who touches this code. Genuinely optional, and I am happy to fill it in from the diff if you would rather not.

If you would like to add the `db.ts` line and the template notes yourself, go for it and I will review it quickly. If you would rather not, just say the word and I am happy to do both and land it with you credited as the author, since the design and all the hard parts are yours either way. No pressure at all.

CI is green, so the branch itself is in good shape.

Really nice work. 🙏

---

PR #288, opened 2026-07-25
feat(control): open a new chat from an external controller

Yoshiofthewire:
Carved down to `new`, as you suggested. `compact` comes back on its own once the blank case has somewhere to render.

## What does this PR do?

Extends `POST /control` with a `chat` command and one verb, `new`, so an external controller can open a fresh chat tab — the chat action that had no equivalent on the bridge. Stage 2 of the Stream Deck integration, on top of #237.

Three changes from what you reviewed:

**`compact` is gone from the closed set.** Not documented-around — removed from `CHAT_DO`, from `ControlCmd`, and from the panel, so `{"cmd":"chat","do":"compact"}` is now a `400` like any other unknown verb. Which also means the route's own justification stands as written: opening a tab spawns nothing, so `control.ts` and `index.ts` still say only what is true, and `docs/EXTENDING.md`'s "changes only what is *shown*" needed no edit.

**The mailbox lapses after 30s.** You were right that a broadcast plus module state with no expiry is a command waiting to fire at the wrong moment. `pending` now carries the time it was latched, and both readers go through one staleness check — a `peek` that outlived a `take` would have the seeding effect defer to a command that never runs, and open the panel on nothing at all. 30s covers the panel opening and waiting on its repo list, which is the whole reason the mailbox exists; past that, a press is not a request for a chat now.

**The seeding effect stands down for a pending `new`**, via the non-draining `peekChatIntent()` you described — for the reason you gave, that the drain cannot see the tab seeding just made. I checked the failure mode you worried about, where seeding defers and then nothing drains: it cannot happen here, because seeding only defers when `defaultCwd` is set, and that is the same fallback `add()` resolves, so the `add()` it deferred to is the one that cannot refuse.

One correction, since it may matter for the follow-up: the compact turn is built client-side at `ChatPanel.tsx` — it calls the ordinary `send()` with the literal `"/compact"`, and the server special-cases nothing. `chat.ts:220-232` is the generic argv builder for every turn. So the blank-bubble fix lives entirely in the stream-frame parsing in `chatStore.ts`, which I think agrees with where you were pointing.

## How was it tested?

- New `web/test/chat-intent.test.ts` — 8 cases pinning the mailbox: latch/take round-trip, take clears and peek does not, one slot rather than a queue, a lapsed command reading null from *both* readers, one that survives a realistic wait, and subscriber add/remove including a throwing listener.
- `server/test/control.test.ts` now pins a closed set of one: `"compact"` joins `""`, `"send"`, `"/compact"`, `"delete"`, a number, `null`, an object and a missing `do` in the rejection list.
- Full suite green: **908 pass, 0 fail** (290 `web` + 618 `server`). `bunx tsc --noEmit` clean in both halves.
- Exercised end-to-end against a real server with a stub controller: `{"do":"new"}` → `200` and the workspace opens on chat; `{"do":"compact"}` and `{"do":"send"}` → `400`.
- For the double tab specifically, on a cleared profile with the workspace shut: **one** tab, and with the `peek` guard removed **two** — so the guard is load-bearing, not decoration.


SirAllap:
Thank you for this — and the reasoning in `chatIntent.ts` is the best part of it. You are right that `controlBus` cannot carry this one: `Workspace.tsx:83` gates the entire view tree on `{open && …}`, so `ChatView` genuinely does not exist at the moment the command that opens it arrives, and a subscription made a render later would drop exactly the press that matters most. One slot rather than a queue, drain on mount *and* subscribe so both orderings work with no timing assumption, and the verb as a closed set on the server with a comment saying why it is closed — that is the shape I would have wanted. The `sessionId` and `sending` guards show you thought about what a real turn costs, too.

- `bun test` — 273 pass in `web/`, 617 in `server/` (890, as you said)
- `bunx tsc --noEmit` clean in both halves
- CI green across all six checks

I do want to hold this one a little though, and it is `compact` rather than `new` that is the problem. Three things, and the first I would not have found without running it.

**A successful `/compact` produces no visible output at all.** I built the exact argv `server/src/chat.ts:220-232` builds and ran it against Claude Code 2.1.220, on a session with enough history to actually compact:

```
system  / init
system  / compact_boundary
user      content='This session is being continued from a previous conversation…'
user      content='<local-command-stdout>Compacted </local-command-stdout>'
result  / success    result=''
```

There is no `assistant` frame anywhere in that stream. `chatStore.ts:532` pushes `{ role: "assistant", text: "", streaming: true }` the moment a turn is sent, and only `t === "assistant"` ever fills it (`chatStore.ts:568`), so the success path renders as a `/compact` user bubble followed by a permanently blank assistant bubble, with the `finally` merely flipping `streaming` off. The cruel part is the other direction: on a session too short to compact, the CLI *does* emit an assistant frame — I captured `'Not enough messages to compact.'` — so the only case that renders any text is the one that did not work. Press the button on a real session and it is indistinguishable from a hang.

**The mailbox has no expiry, and `/control` is a broadcast.** `index.ts:691` sends the frame to every connected client, and the drain returns *before* `subscribeChatIntent` while `reposKnown`/`scopeKnown` are false. So any client whose workspace happens to be closed — a second browser tab, the desktop window sitting behind the dashboard — latches `pending` and never drains it, and `pending` is module state that no unmount clears. The next time somebody opens the chat view on that client, for any reason at all, a `/compact` nobody asked for fires against whatever session is active. And when the workspaces *are* all open, N dashboards each fire their own, so two `claude --resume` processes append to one transcript — the second-writer corruption that `ChatPanel.tsx:571-574` and `chatStore.ts:473-477` both go out of their way to prevent.

**The two empty tabs are deterministic rather than a race, and the obvious guard does not fix them.** The seeding effect at `ChatPanel.tsx:509` is declared a hundred lines before the intent effect, so when `reposKnown` and `scopeKnown` land it runs first in the same commit, creates the tab, and then `add()` creates the second. I first reached for the obvious fix — skip `add()` when the active chat is pristine — and it does not work: `activeIdRef.current` is assigned during render (`:482`), so in that same passive-effect pass it is still `""` and the guard cannot see the tab that was just seeded. It also silently breaks the legitimate case, because parking on an untouched tab is exactly where the seed leaves you, and the intent has already been drained by then so the button just dies with no feedback. The shape that does work is the other way round: a non-draining `peekChatIntent()`, and have the seeding effect stand down when a `"new"` is pending, so `add()` produces the one and only tab.

Two smaller things while I am here. `docs/EXTENDING.md:122-131` presents that table as the authoritative closed set, and `README.md` enumerates the route as "switch view, toggle workspace, theme, zoom" — neither knows about `chat`, and `compact` is the one verb that really needs documenting, since unlike the others it spends money. And the route's own justification needs a trim: `control.ts:19-21` still says "Nothing here executes", and `index.ts:679-684` says it "changes only what is *shown*". The capability claim survives — you can type `/compact` into the composer yourself, so this grants nothing the keyboard does not — but "changes only what is shown" stops being true the moment a command spawns a `claude` subprocess, and your own new comment three lines below says so.

The cleanest way through, if you like it: split `new` off and land it on its own. It has none of these problems, it is the verb someone will actually bind to a physical button, and the mailbox earns its keep on that alone. Then let `compact` come back separately once the blank case has somewhere to render — `compact_boundary` looks like a good hook, and the summary does arrive, just as a `user` frame.

Happy for you to carve it up that way and I will review it fast. If you would rather not, say the word and I will do the split myself with you credited as the author, since the mailbox design and all the hard thinking here are yours either way. No rush on any of it, and nothing here is a criticism of the idea — it is a good one, and `new` I would take today.

🙏

---

SirAllap (approved):
Approved — this answers all three, and answers them better than what was asked for.

**`compact` is properly gone.** Removing it from `CHAT_DO`, `ControlCmd` and the panel rather than documenting around it is the right call, and it means the route's justification stays literally true instead of nearly true. That distinction matters more here than it looks: `/control`'s whole claim is that it grants no capability the keyboard doesn't, and a closed set of one is the cheapest way to keep that honest.

**The TTL, and specifically `fresh()` being shared.** A `peek` that outlived a `take` would have the seeding effect defer to a command that never runs and open the panel on nothing — and the test that pins it (`a command nobody drained lapses, to both readers alike`) asserts *both* readers, which is exactly where that bug would have hidden. Good instinct to test the invariant rather than the function.

**The seeding stand-down.** The comment explaining why `peek` and not `take` — that the drain is the one that acts on it and cannot see the tab seeding just made because `activeIdRef` is assigned during render — is the kind of note that saves the next person an hour. And the reasoning about why "seeding defers, nothing drains" cannot happen holds: seeding only defers when `defaultCwd` is set, which is the same fallback `add()` resolves, so it defers to an `add()` that cannot refuse. I walked the self-healing case too — if `add()` did fail, the intent is spent, `chats.length` stays 0, and the next seeding pass peeks null and seeds normally. It recovers.

Verifying the guard by removing it and counting tabs — one with, two without — is how that should be demonstrated. Thank you for doing it that way rather than asserting it.

And thank you for the correction on where the compact turn is built. `ChatPanel.tsx` calling the ordinary `send()` with a literal `"/compact"` puts the blank-bubble fix in `chatStore.ts`'s stream-frame parsing, which is a different place than we were pointing. That saved the follow-up from starting in the wrong file.

CI green across all six. `chatIntent.ts` reads like it belongs in this codebase — the module doc explains why a mailbox and not a bus, which is the question anyone opening that file would have.

---

PR #289, opened 2026-07-25
fix(models): track the Claude 5 family, and the 1M context window

Yoshiofthewire:
The model picker, the chat default and both demo seeders still named Claude Opus 4.8. That model is **current, not retired**, so nothing was broken — but the list had fallen a generation behind what the CLI actually runs. Adds Fable 5, moves the default to Opus 5, keeps the 1M variant entry alongside it.

## The part that was actually wrong

`contextWindow.ts` → `familyWindow()` returned `200_000` for every Claude id, with a comment asserting that was *"every current Claude model, absent a suffix"*. That held a generation ago. Today only Haiku is 200k — the 5 family and Opus/Sonnet 4.6 onward all ship 1M.

This is the number the Radar's compaction ring (`derive.ts` → `a.ctxLimit`) and the chat context meter are drawn against, so the effect wasn't an error but a **mis-draw**: an Opus 5 session read as *about to compact* at roughly a fifth of its real capacity. The file's own promote-on-observation rule blunted it — but only after a prompt passed 200k, and only to the next rung up (400k), so the reading stayed wrong, just less wrong.

Moving the default to Opus 5 is what would have made this bite by default, which is why it belongs in this PR rather than a separate one.

Worth noting the irony: the file's header comment describes this exact class of silent wrongness as the reason it was written.

## Two deliberate choices in the fix

**The 1M set is the wide list, not the narrow one.** An id nobody has taught the file about falls to 200k. That's the recoverable direction — under-reading is corrected by the observation rule; over-reading has no backstop and would quietly draw a session that really is about to compact as having plenty of room.

**`-4-[678]`, not a general `4-\d`.** The cut lands mid-generation: Opus 4.5 is 200k, Opus 4.6 is 1M. `web/test/context-window.test.ts` pins that boundary explicitly, along with unknown-id fallback, observation promotion, and suffix-wins.

## Also

- `server/src/chat.ts` dropped to Opus 4.8 when a request carried no model or a malformed one, while `DEFAULT_MODEL` had moved on. Only reachable on a bad request, but the two shouldn't disagree about what "default" means. Coupled by comment, matching the existing convention in `control.ts` ("Kept in step with shared/types.ts by hand") — happy to promote it to a shared constant instead if you'd rather it were enforced.
- `hooks/seed_demo.py` was still seeding Opus 4.8 while `web/src/lib/demo.ts` had been updated — same job, half done.

**Pricing needed no change.** `PRICE_TABLE` matches on family substring (`opus`/`sonnet`/`haiku`/`fable`), so Opus 5 and Fable 5 already resolve correctly, and the `fable` row was already there. Test fixtures using `claude-opus-4-8` as opaque strings are left alone — swapping them is diff noise with no behaviour change.

## Testing

- 9 new tests in `web/test/context-window.test.ts`.
- Full suite green: **897 pass, 0 fail** (888 on `main` + 9). `web` typecheck clean, `seed_demo.py` compiles.

## Ordering

Independent of #288 — different concern, and the two touch different hunks of `ChatPanel.tsx`, so they should merge in either order without conflict. If #288 lands first this may want a trivial rebase.

### DIFF
--- a/hooks/seed_demo.py
+++ b/hooks/seed_demo.py
@@ -29,7 +29,7 @@ def _agentglass_local_only(url):
         sys.exit(0)
 
 APPS = ["api-refactor", "docs-agent", "test-writer", "migration"]
-MODELS = ["claude-opus-4-8", "claude-sonnet-5", "claude-haiku-4-5"]
+MODELS = ["claude-opus-5", "claude-sonnet-5", "claude-haiku-4-5"]
 TOOLS = ["Bash", "Read", "Edit", "Grep", "Write", "WebFetch", "Task"]
 _seed = random.Random(7)
 
--- a/server/src/chat.ts
+++ b/server/src/chat.ts
@@ -212,7 +212,9 @@ export function chatStream(cwd: unknown, message: unknown, model: unknown, resum
   // An image on its own is a complete thought ("what's wrong with this?"), so a
   // turn only needs text when it carries nothing else.
   if (!message.trim() && !imgs.length) return err("invalid message");
-  const m = typeof model === "string" && MODEL_RE.test(model) ? model : "claude-opus-4-8";
+  // Keep in step with DEFAULT_MODEL in web/src/lib/chatStore.ts — this is the
+  // same default, reached when a caller sends no model or a malformed one.
+  const m = typeof model === "string" && MODEL_RE.test(model) ? model : "claude-opus-5";
   let pm = typeof mode === "string" && MODES.has(mode) ? mode : "default";
   if (pm === "bypassPermissions" && !BYPASS_ALLOWED) pm = "default"; // opt-in only
   const rid = typeof resumeId === "string" && SESSION_RE.test(resumeId) ? resumeId : "";
--- a/web/src/components/ChatPanel.tsx
+++ b/web/src/components/ChatPanel.tsx
@@ -36,8 +36,9 @@ import { SidebarGrip } from "./SidebarGrip.tsx";
 // list is its own job. The 1M entry is here because it is what this fix buys:
 // the suffix now survives the server, so the window is actually selectable.
 const MODELS = [
-  { id: "claude-opus-4-8", label: "Opus 4.8" },
-  { id: "claude-opus-4-8[1m]", label: "Opus 4.8 · 1M" },
+  { id: "claude-fable-5", label: "Fable 5" },
+  { id: "claude-opus-5", label: "Opus 5" },
+  { id: "claude-opus-5[1m]", label: "Opus 5 · 1M" },
   { id: "claude-sonnet-5", label: "Sonnet 5" },
   { id: "claude-haiku-4-5", label: "Haiku 4.5" },
 ];
--- a/web/src/lib/chatStore.ts
+++ b/web/src/lib/chatStore.ts
@@ -148,7 +148,7 @@ const chats = new Map<string, Chat>();
 const subs = new Set<() => void>();
 let seq = 0;
 
-export const DEFAULT_MODEL = "claude-opus-4-8";
+export const DEFAULT_MODEL = "claude-opus-5";
 export const DEFAULT_MODE = "default";
 
 // A cached snapshot, rebuilt only when something actually changes.
--- a/web/src/lib/contextWindow.ts
+++ b/web/src/lib/contextWindow.ts
@@ -25,12 +25,31 @@ function suffixWindow(m: string): number | null {
   return hit[2] === "m" ? n * 1_000_000 : n * 1_000;
 }
 
+/**
+ * Claude ids whose default window is 1M: the 5 family (Fable, Mythos, Opus,
+ * Sonnet) and Opus/Sonnet 4.6 onward. Everything else Claude — Haiku, Opus 4.5
+ * and older, Sonnet 4.5 and older, the 3.x line — is 200k.
+ *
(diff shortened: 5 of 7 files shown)


SirAllap:
Thank you for this one — you went after the reason rather than the table, and that is what makes it hold up. `familyWindow()` answering 200k for everything Claude was the worst shape a bug can take here: it never errors, it just draws a confident wrong picture on the two surfaces whose whole job is telling you when you are about to lose context.

I pulled the branch and went looking for evidence against it, and found the opposite. Across the real transcripts under `~/.claude/projects` on this machine there are 2,955 assistant turns carrying usage. Every single one records a **bare** model id — 2,392 × `claude-opus-5`, 531 × `claude-opus-4-8` — and not one carries a `[1m]` suffix, because the transcript format has no field that could hold one. 829 of those turns (28.1%) had a prompt above 200k, the largest being **580,471 tokens on a bare `claude-opus-5`**, and 549,579 on a bare `claude-opus-4-8`. A 580k prompt cannot exist in a 200k window, so on the scanner path — which is essentially all production data, since `suffixWindow` is unreachable there — bare ids routinely *are* 1M sessions, and `familyWindow()` is the only thing that ever gets to set their ceiling.

That inverts what I first assumed the risk here was. Pre-PR, a bare `claude-opus-5` sitting at 150k rendered `150.0k / 200.0k · 75%` — an about-to-compact reading for a session that was 15% full — and the promotion rule could only rescue it *after* the bar had already shown a false 100%. So this is not a widening of an adequate default; it removes a systematic under-statement that the promotion rule could only ever paper over after the fact. Writing the regex as the wide list so an unknown id falls to 200k is the right instinct for exactly the reason you give in the comment.

The tests are the part I liked most. Pinning `claude-opus-4-5` and `claude-sonnet-4-5` at 200k right next to `4-6` at 1M documents why the pattern cannot be a loose `4-\d` better than the comment does on its own, and `ctxLimitOf("claude-something-7", 250_000) === 400k` pins the self-healing rule itself rather than just the table.

- `bun test` — 282 pass in `web/`, 615 in `server/` (897, as you said)
- `bunx tsc --noEmit` clean in both halves
- `bunx vite build` green

**One thing got missed, and it is a second table rather than anything wrong with yours.** `web/src/lib/demo.ts:115` keeps its own copy of the ceiling:

```ts
const limit = s.model.includes("gemini") ? 1_000_000 : s.model.includes("gpt-5") ? 400_000 : s.model.includes("gpt") ? 128_000 : 200_000;
```

So the fabricated context for the Claude demo sessions still grows against 200k and snaps back at `limit * 0.92` = 184,000, while `derive.ts:307` now measures it against 1,000,000. I probed it on your branch: those blips top out at 18% of the ring and sit in the inner third permanently, so the drift toward the edge and the snap back — which the comment three lines above calls "the point of it" — stops happening in the demo build. Since `demo.ts` already mirrors the same gemini / gpt-5 / gpt cascade, calling `ctxLimitOf(s.model)` there would delete the duplicate rather than update it.

**Two comments are now stale, and the load-bearing one is mine.** `server/src/chat.ts:38-42`, three lines above the default you change, still says the suffix is how Claude Code asks for "a non-default context size", and that without it a request "silently fell back to the 200k default". That was true when I wrote it — the CLI gated 1M behind the `context-1m-2025-08-07` beta header, and the family test was limited to `claude-sonnet-4` and `opus-4-6` — and it is not true of the 5 family, where 1M is simply the window. Leaving it in place means the repo asserts both propositions at once, which is exactly the thing that costs the next reader an afternoon.

`ChatPanel.tsx:35-37` has the same problem from the other end. With `opus-5` in `CLAUDE_1M`, `ctxLimitOf("claude-opus-5")` and `ctxLimitOf("claude-opus-5[1m]")` both return 1,000,000 — I probed both — so the "Opus 5 · 1M" entry is now a choice with no observable effect, sitting under a comment that says it exists because the window is "actually selectable".

One smaller note, take it or leave it: `contextWindow.ts` now knows `mythos` and the test asserts it, but `pricing.ts` has no `mythos` row, so a Mythos session falls through to `DEFAULT_PRICE` at $3/$15 rather than $10/$50. Nothing breaks — it just means one table learned about a model and the other did not.

None of that is blocking, and none of it is a reason for you to push another round. Happy for you to add the `demo.ts` line and the two comment fixes if you feel like it, and I will review it quickly. Equally happy to merge this as-is and land them myself in a follow-up, since the diagnosis and the whole shape of the fix are yours either way. Just say the word.

Really good work on this one. 🙏

---

SirAllap:
Merged. 🎉

I took the `demo.ts` line and the two stale comments off your hands rather than have this wait on another round — they are in #291, and nothing in your diff was touched. What landed on top, so you can see it rather than discover it:

- `demo.ts` now calls `ctxLimitOf` instead of keeping its own cascade, so the demo's Claude blips drift out to 92% of the ring again rather than parking at 18%.
- `chat.ts:38-42` and `ChatPanel.tsx:35-37` now say what is actually true — the suffix is a no-op on the first-party API for the 5 family, and still decides the window behind an LLM gateway or on Bedrock / Vertex / Foundry, which is why `MODEL_RE` keeps letting it through and why the `· 1M` entry is still worth offering.
- `pricing.ts` grew the `mythos` row your test implied, with a case covering both premium tiers.

The 580k number is the one I keep coming back to. That bar had been quietly lying to me for weeks and I had not noticed, because it never looked broken — it just looked alarming. Thanks for going after the reason rather than the symptom. 🙏

---

PR #353, opened 2026-07-27
fix: retire stale sessions in timeline

Stevekaplanai:
## What does this PR do?

The Sessions timeline now uses the existing `sessionIsLive()` rule instead of treating every row without `ended_at` as permanently live.

AgentGlass already retires an open session after two minutes of silence because killed processes cannot always write a terminal event. The timeline bypassed that shared rule, so an abandoned session could keep pulsing forever while other screens correctly treated it as finished.

Found while adapting AgentGlass as the observation layer for [AgentHost](https://agenthost.space)'s supervised Claude/Codex/Hermes room.

## How was it tested?

- `bun test test/session-live.test.ts` — 4 passed
- `bunx tsc --noEmit` — passed
- `bunx vite build` — passed
- Full web suite: 417 passed; two unrelated Windows-environment tests failed (`diff-theme` color variation and `no-native-dialogs` path normalization)


SirAllap:
Merged, thank you.

You found the one screen that was going around the shared rule. `sessionIsLive` already governs the session modal and the chat panel, the timeline just never got updated, so a killed session kept pulsing there long after every other view had retired it. Good catch, and a better one for having come out of actually running the thing rather than reading it.

One thing from your notes: the two tests failing for you are on Windows, which is the platform I have the least real coverage on. If you hit anything odd there, even small, I would like to hear it. Same goes for Hermes and Kimi-K3 whenever you get that far.

This goes out in the next release.

---

PR #357, opened 2026-07-27
feat: make external ingest retry-safe and map Codex usage

Stevekaplanai:
## What does this PR do?

External harnesses can now send an optional `event_id` and `reported_cost_usd` to `POST /ingest`. Keyed retries are transactionally deduplicated, so a timed-out retry cannot duplicate the event, full-text search row, session totals, alert, or live update. Sender-reported cost wins for that event, while a separate local-pricing baseline keeps later cumulative transcript costs accurate.

Codex CLI OTLP logs now recognize `event.kind=response.completed` and its native token-count fields. Traces and logs share one OpenTelemetry usage mapper, including the official dotted cache attributes, so cached input is separated instead of charged twice and fully cached turns no longer disappear. Generic `event.kind` logs still require Codex evidence.

The public boundary also rejects malformed identity fields and non-finite or unrealistic costs, keeps internal pricing state out of API/WebSocket session data, and documents UUIDv4 retry keys. Existing unkeyed senders keep their append-only behavior.

Found while adapting AgentGlass as the observation layer for [AgentHost](https://agenthost.space)'s supervised Claude/Codex/Hermes room. This PR contains only provider-neutral ingest and telemetry improvements, with no AgentHost branding or visual customization.

## How was it tested?

- `bun test server/test/db-pricing-baseline-migration.test.ts server/test/ingest.test.ts server/test/ingest-usage.test.ts server/test/ingest-reliability.test.ts server/test/cumulative-cost.test.ts server/test/pricing.test.ts server/test/otlp-cap.test.ts server/test/otlp-codex.test.ts server/test/session-lifecycle.test.ts` - 61 passed
- `bunx tsc --noEmit -p server/tsconfig.json` - passed
- `bun run build` - production build passed (existing large-chunk warning only)
- Three independent read-only QA worktrees reviewed exact commit `64b2a7d` for code correctness, security, and TypeScript/data-model behavior - all passed

### DIFF
--- a/README.md
+++ b/README.md
@@ -824,7 +824,7 @@ cannot be configured by `export` at all.
 
 | Route | Description |
 |---|---|
-| `POST /ingest` | Ingest an event `{source_app, session_id, hook_event_type, event_id?, reported_cost_usd?, payload?, chat?, model_name?}`. `event_id` makes retries idempotent; reported cost overrides estimated cost for that event. |
+| `POST /ingest` | Ingest an event `{source_app, session_id, hook_event_type, event_id?, reported_cost_usd?, payload?, chat?, model_name?}`. A high-entropy `event_id` makes retries idempotent; reported cost (maximum `$100,000`) overrides estimated cost for that event. |
 | `POST /v1/traces` | OTLP/HTTP (JSON + protobuf) — maps OpenTelemetry `gen_ai.*` spans to events (any provider). |
 | `POST /v1/logs` | OTLP/HTTP (JSON + protobuf) — maps OpenTelemetry GenAI log records to events (e.g. Codex CLI). |
 | `GET /events/recent?limit=` | Latest events. |
--- a/README.md
+++ b/README.md
@@ -821,7 +821,7 @@ cannot be configured by `export` at all.
 
 | Route | Description |
 |---|---|
-| `POST /ingest` | Ingest an event `{source_app, session_id, hook_event_type, payload?, chat?, model_name?}`. |
+| `POST /ingest` | Ingest an event `{source_app, session_id, hook_event_type, event_id?, reported_cost_usd?, payload?, chat?, model_name?}`. `event_id` makes retries idempotent; reported cost overrides estimated cost for that event. |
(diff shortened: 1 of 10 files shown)


SirAllap:
Merged. This is careful work, and the idempotency half is exactly right.

Three things worth naming, because they are the parts that are easy to get wrong and you got right. The migration checks `PRAGMA table_info` instead of catching a duplicate-column error, and wraps the ALTER and the backfill in one transaction, so a crash cannot leave the column there with every baseline still zero. The idempotency index is partial, so every existing unkeyed sender carries on exactly as before. And `if (!result.inserted) return result` sits above the cache invalidation, the WebSocket frame, the open-tool push and the alert, which is the whole point of the change and the one place a retry would otherwise do visible damage.

I ran the full server suite here: 740 pass, 0 fail.

I also went and checked the token mapper against the spec, because subtracting cache tokens from the input total looked wrong to me at first glance and I wanted to be sure before merging. You are right and I was wrong: semantic-conventions v1.40.0 defines `gen_ai.usage.input_tokens` as inclusive of cached tokens, and the Anthropic section specifically instructs instrumentations to sum the provider's exclusive values to produce it. So the subtraction is the required normalisation back to the disjoint form `pricing.ts` charges, not a bug, and the comment on `otlp.ts:276` is accurate. Reading the official dotted keys also fixes real double-charging for anything instrumented with openllmetry, which is most of the Anthropic ecosystem.

One edge worth a follow-up rather than a change here. OpenLIT's Anthropic instrumentation sets `gen_ai.usage.input_tokens` to Anthropic's raw exclusive value while also emitting the dotted cache attributes on the same span, which goes against the spec's MUST. Against that shape the subtraction clamps input to zero on any cached session, and the `Math.max(0, ...)` makes it silent rather than loud. Not your bug to carry, but if you ever want a cheap guard, treating a total smaller than the cache attributes as already-exclusive would catch it.

The Windows offer still stands. And thank you for keeping this provider-neutral without being asked, that is the reason it was an easy merge.

---

PR #361, opened 2026-07-27
feat(ingest): add Kimi K3 compatibility

Stevekaplanai:
## What does this PR do?

Adds first-class Kimi K3 ingestion support without introducing any AgentHost branding or customization.

- recognizes `k3`, `kimi-k3`, `kimi-code/k3`, and Moonshot model names as **Moonshot / K3** in both server and web provider buckets
- supports Kimi/Moonshot's direct `cached_tokens` field and OpenAI-compatible `prompt_tokens_details.cached_tokens`
- subtracts cache hits from `prompt_tokens` before pricing, preventing the same tokens from being charged as both ordinary input and cached input
- adds K3 API pricing and preserves `reported_cost_usd` as the authoritative value when a caller knows the exact charge
- extends the bundled Python hook forwarder with a `--model-name` fallback for Kimi Code hook payloads
- documents ready-to-copy Kimi Code `config.toml` hook blocks and external usage ingestion

## Why the cache split matters

Kimi's OpenAI-compatible usage shape includes cached tokens inside `prompt_tokens`. For a response reporting 1,200 prompt tokens and 800 cached tokens, AgentGlass now records:

- 400 ordinary input tokens
- 800 cache-read tokens

That keeps token totals and cost charts accurate instead of pricing 2,000 input-side tokens.

## Hook scope

Kimi Code hooks provide the live session and tool lifecycle. Their hook payload does not currently include token usage, so exact token and cost charts still require a Kimi/Moonshot adapter to send the final `usage` object to `POST /ingest`, or to supply `reported_cost_usd`.

## Pricing source

K3 rates follow Kimi's published API pricing:
https://www.kimi.com/help/kimi-api/api-pricing

- input / cache miss: $3 per million tokens
- cache hit: $0.30 per million tokens
- output: $15 per million tokens

## How was it tested?

- focused ingest, pricing, provider, cumulative-cost, reliability, lifecycle, and real-hook suite: **99 passed, 0 failed, 216 assertions**
- real Python hook process posts an official Kimi-shaped event to a live loopback Bun receiver
- production web build passes with `bun run build`
- regression coverage confirms an unrelated model such as `task3-custom-model` does not accidentally match exact `k3`
- upstream CI build, JavaScript/TypeScript CodeQL, Python CodeQL, and GitGuardian checks pass on commit `91181e3`

## Files of interest

- `hooks/send_event.py`
- `server/src/ingest.ts`
- `server/src/pricing.ts`
- `server/src/db.ts`
- `web/src/lib/format.ts`
- `docs/EXTENDING.md`
- `server/test/kimi-hook.test.ts`

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,11 +699,12 @@ proceeding, and remember agentglass being down then blocks every gated call.
 
 ### Kimi Code CLI and Kimi K3
 
-Kimi Code CLI can stream its session and tool lifecycle through the bundled
-hook forwarder. Agentglass recognizes the real K3 names (`k3`, `kimi-k3`, and
-`kimi-code/k3`) as **Moonshot / K3**, understands both Moonshot cache-token
-formats, and applies K3's input, output, and cache-hit rates without counting
-cached prompt tokens twice.
+[Kimi Code CLI hooks](https://moonshotai.github.io/kimi-code/en/customization/hooks)
+can stream its session and tool lifecycle through the bundled hook forwarder.
+Agentglass recognizes the real K3 names (`k3`, `kimi-k3`, and `kimi-code/k3`)
+as **Moonshot / K3**, understands both Moonshot cache-token formats, and applies
+K3's input, output, and cache-hit rates without counting cached prompt tokens
+twice.
 
 Kimi's hook payload does not carry token usage, so hooks populate the live
 session/tool views; a Kimi/Moonshot adapter can send its final `usage` object to
--- a/README.md
+++ b/README.md
@@ -697,6 +697,21 @@ proceeding, and remember agentglass being down then blocks every gated call.
 
 ## Any provider — via OpenTelemetry (OpenAI, Gemini, Bedrock, …)
 
+### Kimi Code CLI and Kimi K3
+
+Kimi Code CLI can stream its session and tool lifecycle through the bundled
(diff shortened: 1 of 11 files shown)


SirAllap:
Merged, thank you. I pulled the branch and ran it here before it landed: 751 server tests green, web typecheck and build fine.

The part I want to call out is the restraint in `usageFrom`. The split only fires when `input_tokens` is absent, so Anthropic's already-disjoint shape passes through untouched, and a fully cached turn reporting `input_tokens: 0` stays 0 instead of getting rewritten. That was the case I went looking for, and you'd already handled it. Second PR in a row where the careful half was done before I asked.

I also checked the rates against a few independent trackers rather than take them on faith, and they hold: $3 in, $15 out, $0.30 on cache hits, flat across the whole 1M window.

One small thing I folded in myself rather than send back to you, since it came out of that last detail. `providerOf` and `modelLabelOf` both accept any `k3[...]` suffix, but the price table listed the windows one at a time, so `k3[512k]` read as Moonshot / K3 in the UI while being priced by the fallback behind it. Since the rate is flat across the window anyway, `exact` now takes the suffix the way the other two do. #363.

Still keen to see Hermes whenever it happens, and no rush on it.

---

Stevekaplanai:
Hermes is in. I’m almost setup from my phone to remote in Codex built this
btw.

The whole thing is ready I just have to reboot. But I’m remote. So let’s
see what happens!

PR #451, opened 2026-07-31
Recognize MiniMax as a provider and price M3/M2.7

octo-patch:
Reason: MiniMax events fell back to unknown labels and default pricing instead of the current model-specific rates.

providerOf had no rule for minimax, so MiniMax sessions bucketed under the unknown provider; PRICE_TABLE had no MiniMax-M3 or MiniMax-M2.7 entries, so costUsd fell back to DEFAULT_PRICE. This adds:

- a `minimax` provider branch in shared/models.ts (used by both server and web), with MiniMax M3 / MiniMax M2.7 display labels ahead of a generic MiniMax label;
- two PRICE_TABLE rows in server/src/pricing.ts with the current per-MTok input/output/cache-read/cache-write rates for each id (M2.7 ordered before M3);
- tests covering the matched rates, the per-class cost math, the provider/label parity between server and web, and the M2.7-before-M3 ordering.

Checks run: `cd server && bun test pricing.test.ts provider-parity.test.ts` (109 pass, 0 fail).

### DIFF
--- a/server/src/pricing.ts
+++ b/server/src/pricing.ts
@@ -110,6 +110,13 @@ export const PRICE_TABLE: ModelPrice[] = [
   // Flat across the whole 1M window — there is no long-context tier to add.
   { match: ["kimi-code/k3", "kimi-k3", "moonshot/k3"], exact: ["k3"], label: "K3", input: 3, output: 15, cache_write: 3, cache_read: 0.3 },
 
+  // --- MiniMax ---
+  // M3 keeps cache creation at the input rate (cache_write === input); M2.7
+  // charges cache creation at 1.25x input. Both discount cache reads to a tenth
+  // of input. Order matters: M2.7 is matched before the generic M3 row.
+  { match: ["minimax-m2.7"], exact: ["MiniMax-M2.7"], label: "MiniMax M2.7", input: 0.3, output: 1.2, cache_write: 0.375, cache_read: 0.06 },
+  { match: ["minimax-m3"], exact: ["MiniMax-M3"], label: "MiniMax M3", input: 0.6, output: 2.4, cache_write: 0.6, cache_read: 0.12 },
+
   // --- others (approx) ---
   { match: ["deepseek"], label: "DeepSeek", input: 0.27, output: 1.1, cache_write: 0, cache_read: 0.07 },
   { match: ["grok"], label: "Grok", input: 2, output: 10, cache_write: 0, cache_read: 0 },
--- a/server/test/pricing.test.ts
+++ b/server/test/pricing.test.ts
@@ -100,6 +100,43 @@ describe("costUsd", () => {
     expect(cost).toBe(DEFAULT_PRICE.input);
   });
 
+  test("MiniMax M3 and M2.7 use their own rates instead of the fallback", () => {
+    // The model ids are short and do not share the Gemini/OpenAI collision
+    // problem, but a bare `minimax` substring would otherwise fall through to
+    // DEFAULT_PRICE — the Sonnet-tier $3/$15 — and over-report cost ~5x.
+    expect(priceFor("MiniMax-M3")).toMatchObject({
+      label: "MiniMax M3",
+      input: 0.6,
+      output: 2.4,
+      cache_write: 0.6,
+      cache_read: 0.12,
+    });
+    expect(priceFor("MiniMax-M2.7")).toMatchObject({
+      label: "MiniMax M2.7",
+      input: 0.3,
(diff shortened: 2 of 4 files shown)


SirAllap:
Thanks for this, and welcome. This is a well put together first PR: the provider branch is in the right file, the label list keeps the generic `minimax` entry last where it belongs, and adding the two ids to the parity test list is exactly the habit that file exists to encourage. I checked out the branch and ran the suite, and I get your 109 pass / 0 fail too.

I did dig into the rates before merging, since this table is the one place in the app where a wrong number turns into a confident-looking dollar figure. A few things came back, one of them blocking.

**1. The M3 rates are the long-context tier, not the standard one (blocking)**

MiniMax prices M3 in two tiers. At or below 512K input tokens it is $0.30 in / $1.20 out with cache reads at $0.06. Above 512K, everything doubles: $0.60 / $2.40, reads $0.12. The row has the second set, so a normal agent turn (nowhere near 512K) would report twice the money it actually cost.

Two sources agreed on both tiers:
- https://developer.puter.com/tutorials/minimax-api-pricing/
- https://devtk.ai/en/models/minimax-m3/

Easy trap: most aggregator pages lead with the long-context number, and it does look like the headline rate.

Since the table has no way to express a tier (see the K3 comment right above your rows, which notes K3 has no tier to add), I think the honest fix is the standard rates plus a comment saying a >512K request is under-reported by half. Happy to hear a better idea.

**2. `cache_write: 0.6` for M3 is not a published number**

MiniMax publishes a separate cache creation rate for M2.7 ($0.375, which matches your row) but not for M3. In that case creation bills as ordinary input, so `cache_write === input` says it truthfully, the same way the K3 row does. Worth avoiding `0` here: in this table a zero means the provider does not charge, which is a stronger claim than the sources support.

**3. The comment says a tenth, the numbers say a fifth**

`0.06 / 0.3 = 0.2`, and the same for M3. The rates are right and the sentence is wrong, which is the direction that tends to survive review and mislead the next person.

**4. `exact` never fires with those values**

`priceFor` lowercases the model name before comparing, so `exact: ["MiniMax-M2.7"]` cannot ever equal `m`. Compare the K3 row, which uses lowercase `"k3"`. Nothing breaks because your `match` substrings already cover both ids, which is also why the tests pass with it dead. Since these ids are long enough to be safe as substrings, I would just drop `exact` from both rows.

**5. The ordering property is not load-bearing**

`"minimax-m2.7".includes("minimax-m3")` is `false`, so the two rows match independently of each other. I swapped them and all 109 tests still pass, including the `not.toBe("MiniMax M3")` assertion, so that one is not pinning what the comment says it pins. Something like asserting the rows differ in cache creation would actually catch a mis-resolved id.

**One suggestion, take it or leave it**

MiniMax ships models these two rows do not cover, and `providerOf` now recognises all of them, which is exactly how a model picks up a cost it never earned. A test that `hasPrice("MiniMax-M2")` stays `false` would pin the honest behaviour your change already has.

**Housekeeping**

I approved the workflow runs (first-time contributor, so CI was sitting on `action_required`). I also, too eagerly, pushed a commit with these fixes onto your branch and then reverted it, so if you got a notification about `f0ca67b`, that was me and it is gone. Your branch is untouched at `0f21b0c`. My mistake: the fixes are yours to make, and it is your PR.

If you would rather not chase the rates yourself, say the word and I will push the patch. Otherwise take your time, and thanks again for catching that MiniMax was falling through to the unknown bucket in the first place. That part was a real gap.

SirAllap:
Merged — thanks for this. MiniMax was missing from the provider list entirely, and the `providerOf` regex, the display labels and the parity test were all exactly right.

I adjusted the rates in a follow-up (#461) rather than sending it back, so no action needed from you. For the record of what changed and why: `MiniMax-M3` came in at `$0.60` in / `$2.40` out with a `$0.12` cache read, which is the doubled rate — M3 bills the whole request at 2x past 512K input tokens, and `$0.60/$2.40` is also the price of the `-highspeed` M2.7 variant, so those two numbers sit right next to these models in the docs. Standard is `$0.30/$1.20` with a `$0.06` cache read, matching M2.7. Your M2.7 row was already correct and is untouched.

I also lower-cased the `exact` entries — `priceFor` compares against a lower-cased id, so `"MiniMax-M3"` there never fired — and added a catch-all `minimax` row, because `shared/models.ts` labels any `minimax` id as MiniMax while the price table knew only two, which would have billed the vendor's next model at `DEFAULT_PRICE`. Your own test comment flagged that risk; it just needed the row to go with it.

---

PR #454, opened 2026-07-31
Drive Codex and Google Antigravity from the chat panel

Yoshiofthewire:
The chat panel gains two more agents. It drove `claude`; it now also drives **`codex`** and **`agy`** (Google Antigravity).

All three are driven the same way — spawn the CLI non-interactively, stream its JSONL back — and all three land in the same `ChatMsg` / `ChatTool` / `ChatUsage` shapes, which is what lets one panel render any of them without a branch below the store.

> **Antigravity is not the Gemini CLI, and does not replace it.** Separate binary, separate state, and a model list that offers `claude-sonnet-4-6` and `gpt-oss-120b-medium` alongside Google's. Gemini CLI keeps its existing OpenTelemetry route onto the radar, untouched.

## Each CLI keeps its own vocabulary

Rather than being mapped onto Claude's:

- **Codex** modes are filesystem sandboxes (Read-only / Write in this repo / ⚡ Full access). No allowlist box — it decides at the filesystem boundary, not per tool call.
- **Antigravity** modes *do* land on Claude's four (Ask / Plan / Auto-accept edits / ⚡ Bypass), because it really does decide per call and really does have a plan mode. That is the CLI's shape, not a mapping imposed here.
- Model lists come from each CLI — Codex's `models_cache.json`, `agy models` — so they follow the vendor instead of a table here that goes stale on every release.

A chat is bound to one CLI for life: the agent picker is switchable until it holds a thread, then frozen. A resume id means something only to the binary that minted it.

## Antigravity reports to nobody, so the panel reports for it

Claude reaches the fleet through hooks and Codex through OpenTelemetry. Antigravity exports neither, so a chat started here would appear on the radar nowhere at all.

Its frames already carry everything an event needs, so the stream is teed: out to the browser verbatim, and through `frameToEvent` into the same `ingestBody` path the other intakes use.

```
init         → SessionStart      user_input → UserPromptSubmit
tool ACTIVE  → PreToolUse        tool DONE  → PostToolUse
result       → Turn complete
```

`source_app` is `"antigravity"`, which is load-bearing rather than cosmetic: `agy` can run `claude-opus-4-6-thinking`, so the model name is an actively misleading signal for "which CLI is this?"

The tee is server-side because what the fleet knows should not sit behind a tab that can be closed mid-turn. **Only chats started here are covered** — a terminal `agy` stays invisible, and the README says so rather than implying otherwise.

## Capabilities that are absent are left off, not faked

| | Codex | Antigravity |
|---|---|---|
| Resume from disk | ✅ JSONL rollout | ❌ protobuf in SQLite |
| Cost | ✅ priced server-side | ❌ no price on the stream |
| Context meter | ❌ no per-turn prompt size | ❌ no window reported |

Antigravity's conversations are protobuf blobs on an undocumented internal schema; decoding them would break on any CLI update. And `ctxLimitOf` does not know the Gemini 3.x families, so a context bar would fall back to a default ceiling and state a fullness nobody measured.

## Usage semantics differ, and look identical

Codex reports **cumulative thread totals** — assigned. Antigravity reports **per-turn figures** — added. Measured, not inferred: Antigravity's turn 1 reported 31789 input tokens and turn 2 reported 16609, which cannot be a running total. Getting this backwards over-reports spend severalfold and nothing fails loudly.

## The refactor three agents earned

Two agents justified `agent === "codex" ? … : …`; three did not — a missed branch is silent, a Claude default quietly applied to something that is not Claude. `AGENTS` in `web/src/lib/agents.ts` now holds what differs (label, binary, defaults, unattended mode, whether it takes attachments, whether it has a transcript), so a fourth CLI is an entry rather than a search. Its own module because `chatPersist` needs it too, and importing from the store would put a cycle between them.

One operator opt-in (`AGENTGLASS_CHAT_BYPASS`) covers the unattended mode of all three, since it is the same decision — the README previously documented only the Claude half.

## Verification

- **web:** 966 pass, 0 fail · `tsc --noEmit` clean · `vite build` succeeds
- **server:** 1423 pass. The 2 failures are pre-existing on a clean `main` in this environment (one flaky tmux pane test; one test that assumes `codex` is absent from `PATH`, where this machine has it). Verified by stashing and re-running.
- **Driven for real**, not just unit-tested: a live `agy` turn through `POST /antigravity/send`, confirmed landing in the fleet with the right agent, model, tool pairing (`tool_mix: list_dir ×1`) and tokens. That smoke test is what caught the missing `cwd` / `project_path` on synthesized events — without them a chat landed in the store and then showed up nowhere, which is indistinguishable from never having been recorded. Now pinned by a test.

Design notes: `docs/superpowers/specs/2026-07-31-antigravity-agent-design.md`. `docs/EXTENDING.md` gains what the third CLI taught, where "a third CLI would repeat that shape" used to be a prediction.

### DIFF
--- a/.github/workflows/desktop-binaries.yml
+++ b/.github/workflows/desktop-binaries.yml
@@ -65,6 +65,12 @@ jobs:
           mkdir -p electron/staging
           bun build --compile --target=${{ matrix.bunTarget }} server/src/index.ts \
             --outfile "electron/staging/agentglass-server${ext}"
+          # The Claude model catalogue is read from disk at runtime, not
+          # imported, so --compile does not carry it. It goes beside the sidecar
+          # because that is where claudemodels.ts looks in a packaged app
+          # (dirname(process.execPath)). Without this the desktop build falls
+          # back to a single model id and the dropdown looks broken.
+          cp shared/claude-models.json electron/staging/claude-models.json
 
       # Sync the package version to the tag so a v0.2.0 tag ships v0.2.0 files.
       - name: Sync version to the tag
--- a/README.md
+++ b/README.md
@@ -207,17 +207,46 @@ Electron window compositor rather than xterm.
 
 ![terminal panel](.github/assets/terminal.png)
 
-### 💬 Chat — drive Claude sessions from the browser &nbsp;`c`
+### 💬 Chat — drive Claude and Codex sessions from the browser &nbsp;`c`
 
 Multi-chat against your **local `claude` CLI**: pick a repo/worktree, a model,
 and a permission mode (plan → default / acceptEdits → bypass), then converse —
 replies stream in, tool calls appear as chips, and follow-ups resume the same
 session. Sessions you start here show up in the fleet like any other agent.
 
+#### A second agent: OpenAI Codex
+
+The same panel drives **`codex`** as well. When both CLIs are on the machine an
+agent picker appears next to the repo — it is switchable until the chat has a
+thread, then frozen, because a resume id belongs to the CLI that minted it and
+there is no meaning to handing a live conversation from one to the other.
+
+Codex brings its own vocabulary rather than borrowing Claude's. Its modes are
+filesystem sandboxes (**Read-only**, **Write in this repo**, **⚡ Full access**)
+instead of per-tool permissions, so there is no allowlist box for a Codex chat —
+it decides at the filesystem boundary, not per tool call. The model list is read
+from Codex's own `models_cache.json`, so it follows whatever your CLI currently
+offers instead of a table in this repo that goes stale on every release. Pasted
+and dropped images are Claude-only: `codex exec` takes images as file paths
+rather than content blocks, so the panel says so instead of dropping them
+silently.
+
+Install and authenticate the Codex CLI separately, then make sure the `codex`
+executable is on the server's `PATH` when agentglass starts. The chat panel uses
+`codex exec --json` for new turns and `codex exec resume` for follow-ups; it does
+not replace the CLI's own login or configuration. Set `CODEX_HOME` when Codex
+keeps its state somewhere other than `~/.codex` — the model cache and rollout
+history are read from that location. To turn off the Codex integration while
+leaving Claude chat available, set `AGENTGLASS_CODEX_DISABLED=1` before starting
+the server.
+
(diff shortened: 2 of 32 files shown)


SirAllap:
Thank you for this — it's a lot of work and it's good work, so rather than send it back for a rebase I've done the merge here: #462. Your ten commits are in it with your authorship intact.

Nothing was wrong with the branch. It forked at `e898c65`, and 0.8.0 then rewrote the chat subsystem it extends, so the only real problem was elapsed time. Three conflicts, all "both sides added to the same line" — the README's notch/bar wording, the type import in `api.ts`, and the import block in `chatStore.ts`.

What made it worth merging rather than asking you to redo it is that the two halves compose on their own. `send()` came out with main's `effort`, `allowedTools` and `engine` inside the Claude arm and your other two agents on the plain turn; the replay path gates on `AGENTS[agent].hasTranscript` while keeping main's `mergeReplayed` and `liveFrom`. Neither side knew the other existed and both are right.

One gap needed filling, and it's the one your own header in `agents.ts` predicts — *"a missed one is silent — a Claude default quietly applied to something that is not Claude"* — only in reverse. Two Claude-only controls landed on main after you wrote the split, so they had no entry in your table: the effort dial (`--effort` is Claude's flag, and `send()` only passes it there) and the tmux pane engine (`engineFor` read a global preference, so with tmux on a Codex chat advertised a pane in its header that was never going to be opened). Both are `AGENTS` flags now — `hasEffort` and `canPane` — because your table was the right place to put them.

I also want to name the review as easy: argv arrays with no shell, `safeAbs` + `repoRootOf` + `inScope` on the working directory, the anchored `MODEL_RE`/`SESSION_RE` so neither a model nor a resume id can turn into a flag, prompt on stdin under a cap, the unattended mode behind the same opt-in as `bypassPermissions`, and a kill switch each. Mirroring `chat.ts` that closely is what made a 4,000-line PR checkable.

Closing this in favour of #462. Nothing needed from you — and thanks again.

---

SirAllap:
This shipped — #462 is merged, so closing here.

To be clear about what that means for the credit: your ten commits are in `main` with your authorship on every one of them. This is closed because the work landed, not because it was turned down.

---

PR #456, opened 2026-08-01
Fix the About tab, which cannot name its own version outside the desktop app

Yoshiofthewire:
The About tab does not load. In a browser it sits on **"Reading version…" forever**.

`/update/status` is gated to the packaged desktop shell's custom scheme, reads included. The reason is sound — the answer names the clone on disk, the remote the updater pulls from, and the tail of the last update log, none of which a phone on the wifi has any business seeing.

The cost went unnoticed: **the About pane is the only place this build's version is written down, and it reads it from there.** Under `bun run dev`, under single-port mode, and on a paired phone, the request comes back 403 and the pane never resolves.

`web/src/lib/updateStore.ts` already knew about the gate — it stops the background poll on `!IS_DESKTOP` precisely because `/update/*` "can only ever produce a 403 in the console every six hours". The About pane never got the same treatment.

## The gate redacts rather than refusing

Identity — version, commit, built-at — goes to anyone the outer origin gate already admitted. Provenance and the update button stay with the desktop shell.

`/update/run` and `/update/log` are untouched: **reading is not running.** The redacted payload blanks `source` and `origin`, omits the `last` stamp entirely (its tail is raw script output that quotes paths verbatim), reports `available: false`, and carries a sentence saying where updating does work. It skips the `git ls-remote` too — there is no update on offer, so there is nothing a network round trip could inform.

`/update/notes` opens on the same terms and for the same reason: About offers a release-notes button whenever the build descends from a tag, so gating it left a browser holding a button that could only ever fail. Its answer is a published release's text, its tag, and whether it came from the clone or from GitHub — no origin URL, no install path, and `tag` is refused unless it matches a release pattern.

Both relaxed routes still sit behind the outer `localOrigin` gate and the token gate, so this only widens the payload for callers that were already admitted.

## The pane's own bug is the general one

```ts
.catch(() => { if (live) setSt(null); });   // …and `!st` renders "Reading version…"
```

A failed read set state to `null`, which is also exactly what "still loading" looks like. So a refused request rendered as a spinner that never resolved — the two states were indistinguishable. Failure now says so, which also covers the 45-second `ls-remote` ceiling nobody would otherwise wait out.

## Verification

- New `server/test/update-status-origin.test.ts` — 6 tests. It **reproduces the 403 first**, then pins both halves: a browser origin gets the version; the install path and remote never appear in the payload; no update is offered; `/update/run` still returns 403; and the desktop origin still gets the full answer including the log tail.
- `server/test/selfupdate*.test.ts` and `release-notes.test.ts` — 30 passing across the four files that cover this area.
- `web/test/update-store.test.ts`, `whats-new.test.ts` — 21 passing.
- `make typecheck` — clean, both halves.

A stale comment in `updateStore.ts` asserting `/update/*` "can only ever produce a 403" from a browser is corrected in the same change; the background poll stays desktop-only, which is still right, because the badge it feeds is about installing an update a browser cannot install.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -82,7 +82,7 @@ import { privateHost } from "./net.ts";
 import { resolveToken, tokenOk, isIntake, isAuthExempt, callerFor, allowed, scopeNeeded, type Caller } from "./auth.ts";
 import { activeDevices, markSeen, revokeDevice, type Scope } from "./devices.ts";
 import { mintTicket, claimTicket, pending as pendingPairings, acceptTicket, rejectTicket, collect as collectPairing, dropTicket, getTicket, MAX_ATTEMPTS } from "./pairing.ts";
-import { updateStatus, startUpdate, updateLog, releaseNotes } from "./selfupdate.ts";
+import { updateStatus, viewerStatus, startUpdate, updateLog, releaseNotes } from "./selfupdate.ts";
 import { rateOk } from "./ratelimit.ts";
 import { noteClient, noteSocket, isLoopback, isSelf, isBlocked, blockDevice, remoteStatus, tailnetNames, refreshTailscale } from "./remote.ts";
 import { parseWindowMs } from "./params.ts";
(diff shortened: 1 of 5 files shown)


SirAllap:
Merged — thank you. This one was a good catch: the About pane was the only place a build writes down its own version, and it had been sitting on "Reading version…" for anyone not in the desktop shell.

Two things I want to name because they made the review easy. Rebuilding the payload field by field in `viewerStatus()` instead of spreading and blanking is the right call — a field added to `BuildInfo` later now has to be named before it can leave the machine. And separating "no status" from "status hasn't arrived yet" in the pane is what turns the next 403 into something visible rather than another permanent spinner.

I checked the two relaxed routes against what they can now reach: `TAG_RE` is anchored to `v\d+.\d+.\d+`, so nothing user-supplied gets into `refs/tags/` or the GitHub URL, the owner/repo come from the local remote rather than the request, and neither route is in `AUTH_EXEMPT` — a configured `AGENTGLASS_TOKEN` still applies. The provenance stays behind the desktop gate, which was the point.

Nothing needed from you. Thanks for the fix and for the regression test that came with it.

---

PR #458, opened 2026-08-04
fix(hooks): non-blocking event send, stop paying localhost IPv6 penalty

dakristia:
The PostToolUse hook was blocking every tool call for ~4.5s median. Root cause: the hooks default to `http://localhost:4000`, which resolves to `::1` first on some hosts while the server binds IPv4-only — and on this managed Windows box a refused connect costs ~2s instead of <1ms. Python startup was ~3% of the cost.

- Default server is now `http://127.0.0.1:4000` in all hook scripts; the local-only guard also rewrites a `localhost` URL to `127.0.0.1`, so existing installs get the fix without re-running setup.
- The event POST now runs in a detached child process, so the hook never waits on the network. A 30s circuit-breaker marker skips the spawn entirely while the server is known down; the transcript scanner backfills anything missed.

Measured: ~2.2s → ~270ms per hook invocation with the server up, ~300ms with it down.


SirAllap:
@dakristia — you closed this yourself, so this is not a request to reopen it. It is to tell you that four pieces of it shipped, and that the review you did here found things nobody else had.

**Merged in #465**, one commit each, with you credited in every commit message:

- **The `AGENTGLASS_ALLOW_REMOTE` guard.** The one that mattered most. `=0`, `=false`, `=no`, `=off` — every spelling a person reaches for to turn something *off* switched the guard *on*. A repo-local settings.json could ship `{"AGENTGLASS_ALLOW_REMOTE": "0", "AGENTGLASS_SERVER": "https://collector.example"}` and read as though it were disabling remote sending while doing the opposite. Verified against the real hooks before and after: that value did deliver.
- **`|| exit 0` on the installed hook command.** Exit 2 from a PreToolUse hook is "deny", python exits 2 on a missing script, and the path written into `~/.claude/settings.json` outlives the clone — so a `git clean` turned the forwarder into a machine-wide block on every tool call, including the ones needed to undo it. Applied in both installers, since `hooksetup.ts` is a port of your Python one and a test already pins that the two agree.
- **`127.0.0.1` instead of `localhost`**, default and rewrite both, so an install that wrote `localhost` months ago gets it without re-running setup.
- **The OpenCode plugin.** Two changes from your version: it defaults to `127.0.0.1` now (it was still on `localhost`, which was the thing the rest of your branch was fixing), and its tests moved into `server/test/` because CI runs neither a `.test.mjs` nor a `test_*.py`. It sits under `connect:` rather than `setup:`, since this repo's vocabulary already has `setup` meaning the Claude hooks and `connect` meaning somebody else's CLI.

**Filed as issues** from your `SECURITY-AUDIT.md` and `BACKLOG.md`, so they outlive this branch: #468 (unvalidated `AGENTGLASS_WEBHOOK` destination, `ANTHROPIC_BASE_URL`-redirectable walkthrough, no walkthrough kill switch), #469 (write routes on `localOrigin`), #470 (CI resolving its toolchain from npm on every run). All four verified against `main` before filing, and all three credit you.

The rest was either already superseded here or specific to your fork — the GLM pricing in NOK, the `Intility` provider label, the CI removal. Nothing to do about any of that.

For what it is worth: the audit was the useful part. It went looking for exfiltration in a tool that reads full transcripts, concluded there wasn't any, and then wrote down the four places where the *shape* was wrong anyway. That is a harder and more honest piece of work than finding a bug, and it is why three of those issues exist.

Good luck with your fork.

---

PR #464, opened 2026-08-05
Plan quota for every provider, not just Anthropic

Yoshiofthewire:
> **Rebased onto `main` (post-#462).** This began stacked on `Improved-Codex-UX` and was rebased onto `claude/codex-antigravity-rebase` while that was in review; #462 merged in the middle of the rebase, so it now sits directly on `main` and needs nothing merged first. It does depend on that work being in `main`: `server/src/codexusage.ts` imports `codexHome`/`CODEX_ENABLED`/`codexModels` from `server/src/codex.ts`, and `web/src/lib/providerContext.ts` imports `AgentKind`.

> **What the rebase changed.** The branch was written against `Improved-Codex-UX`, where the dashboard lived in `App.tsx` and the workspace had a notch. On today's `main` the dashboard is `DashboardView.tsx` and the notch is `TopBar.tsx`, so both surfaces were ported rather than force-fitted: the Usage box goes into `DashboardView`'s money row, and the focused-agent gauge into `TopBar`, keeping that bar's own affordances — the staleness caption, the dim-on-dashboard `quiet` mode, the `hideUnder` breakpoints. Two consequences worth a reviewer's eye, both called out again where they happen: `TopBar`'s Anthropic-only `rate-limited` chip became a provider-driven **no reading** chip whose tooltip is the provider's own sentence, and `anthropicUsage()` now folds the per-model weekly windows (`scoped`) into `windows`, so the per-model meters `TopBar` draws today survive the switch to the shared shape and appear on the dashboard box and in Stats too. `web/test/usage-staleness.test.ts` — a source-text pin — was updated to the ported call sites instead of being deleted.

The cockpit drives three agents and shows quota for one. `UsageWidget` reads Anthropic's OAuth usage endpoint into two meters in the header — shown only when the provider filter is Anthropic, and only above `2xl`, because the header is the most contended strip in the UI. Codex and Antigravity run out exactly the way Anthropic does, and the app says nothing about either.

Now all three, on three surfaces, and the header widget is gone.

## What each provider can actually tell us

This shaped everything else, so it comes first.

**Anthropic** — live, unchanged. `server/src/usage.ts` still fetches the OAuth usage endpoint, caches for 5 minutes and backs off hard on a 429.

**Codex** — a local file read. The CLI already records the rate-limit snapshot it got from the API into `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`, inside its `token_count` events. No credentials, no network, no endpoint that can throttle us for asking — strictly cheaper than the Anthropic path.

It buys that with staleness: `rate_limits` is only written when a turn runs, so the number is as old as your last Codex conversation. That is why `observedAt` is not decoration — **every surface says how old the reading is** — and why the refresh ping below exists at all.

**Antigravity** — nothing exists to read. Verified on 2026-07-31: `agy` has no usage or quota subcommand, `~/.gemini/antigravity-cli/` holds no quota file, its conversation `.db` files carry no usage fields, and its own log shows a `quota_manager.go` that refreshes into process memory and never writes it down. The only route left would be replaying its authenticated backend call, which is out of scope.

So where it is installed it appears as a **labelled gap** — named, with a sentence saying why (*"Quota not reported."*). A provider that simply drops out of the list reads as a bug in agentglass; one that says why reads as the upstream limitation it is.

## Only the CLIs you actually have

The list is filtered on whether each agent's binary is on the PATH. A machine with only Claude installed sees one row, not three — the earlier version spent two thirds of the panel explaining tools its owner had never installed, and the Codex line invited them to "run a Codex turn" with no `codex` binary to run it with.

The gate is **installed**, not **connected**, and the difference is load-bearing. Connecting an agent in Settings › Agents wires its *events* to this server — hooks for Claude, OTel for Codex. Quota arrives by neither road: Codex's lives in the session files its CLI writes whatever it is pointed at, and Claude's behind the OAuth credentials on disk. So an agent you have deliberately disconnected still has a plan that runs out, and hiding that number would withhold something true about a tool you are still using. A test pins that distinction rather than leaving it to the comment.

Filtering made the empty list reachable, so "no quota-reporting CLI on this machine" is now its own state on both surfaces that have room to explain, naming where to change the answer — rather than a panel with no rows in it, which reads as "you have used nothing".

## Three surfaces, three jobs

| Surface | Shows |
|---|---|
| **Dashboard box** (left of Cost, spanning the money row and the timeline) | Every installed provider: meters, plan, reading age, or the reason there is nothing |
| **Stats modal** | The same rows with room for exact reset times and plan type |
| **Workspace notch** | One gauge — the provider you are actually driving |

"In context" in the notch means the focused chat's agent, falling back to the dashboard's provider filter. That distinction is load-bearing: driving a Codex chat shows Codex's quota even when the dashboard behind it is filtered to Anthropic.

The money row and timeline merge into one 12-column grid so the Usage box can span both, and the columns now line up with the cockpit above them — Usage under Sessions, Cost and Latency under the middle column, Sessions under Alerts.

## The refresh ping spends quota to measure quota

Codex's reading only moves when a turn runs, so the remedy is running one: `POST /usage/codex/refresh` executes a minimal `codex exec` and re-reads the snapshot it leaves behind.

Every decision follows from what it costs:

- **off by default**, and the setting says plainly that it uses a small amount of the quota it measures
- **hourly**, with a 15-minute floor so a habit of hitting ⌘R does not become a stream of billed pings
- **driven by the existing 5-minute poll**, not a second timer — one source of truth about when to spend money
- **single-flighted**, so several open tabs collapse into one turn
- **smallest model available**, and never a hardcoded id: `AGENTGLASS_CODEX_USAGE_MODEL` → the tail of Codex's own priority-sorted list → no `--model` flag at all. A model id that is not on the user's plan is a feature that fails for everyone but its author.

The route is gated by `trustedCaller` — the same tier that already gates `/terminal/pty` — and is absent from `READ_POST`/`ANSWER_POST`, so a paired phone on `read` or `answer` scope cannot spend the user's quota.

## What review caught: the same bug, five times

Worth naming for a reviewer, because it is the shape of every serious defect in this branch:

| Where | What a person saw |
|---|---|
| `anthropicUsage` | `available: true`, no windows, no note → a blank row |
| `usageStore` | `!IS_DEMO` gate → demo mode never fetched, spun forever |
| `UsageBox` | `loaded && !rows` → an empty panel |
| `anthropicUsage` (final review) | every failure reported "sign in", including 429 and offline |
| `panelState` (when filtering made it reachable) | an empty list mapped onto "rows" → a panel with no rows |

Four of the five render **nothing**, which reads as "you have used nothing" rather than "we do not know". The other renders a confident wrong answer, which is worse — it told a correctly signed-in user to sign in whenever Anthropic rate-limited for longer than the 30-minute stale window.

All five are fixed. The panel-state decision is an exported pure function shared by both explaining surfaces, so their fallback strings cannot drift apart, and the two surfaces' wording was verified byte-identical.

## Verification

Re-run after the rebase, on a fresh checkout of this branch:

- **web:** 1387 pass, 0 fail · `make typecheck` clean on both halves · `make build` succeeds
- **server:** 1789 pass, 5 fail — all environment-dependent, and the base fails the same five in the same environment. Proven rather than assumed by running the same suite in a worktree at the commit this branches from: the three tmux ones, `with a pane genuinely running > a pin shows up in the listing`, and `connecting one > refuses an agent this machine does not have` (which assumes `codex` is absent from a `PATH` that has it) fail identically on both. None touches a file this changes.
- **`make smoke`** (production bundle in headless Chrome): app mounts, the rail draws 9 views, ⌘-numbers switch them, no console errors.
- **CI caught one thing the local run could not**, fixed in the last commit here: `server/src/usage.ts` captured the credentials path at import, so whichever of the two suites that now import it loaded first froze that path for the other. Local order let `plan-usage.test.ts` win and pass; CI's order made it read the real path and fail fifteen tests. `credPath()` and `offLimits()` are asked when used, the way `lastGoodPath()` directly above them already was. Verified in both orders.

Design notes: `docs/superpowers/specs/2026-07-31-provider-usage-gauges-design.md`. Both the spec and the implementation plan carry inline amendment notes where the shipped code diverges from what they originally described, so a later reader sees the correction rather than reintroducing what was fixed. The install-filter and the shorter Antigravity wording came after those documents and are described here rather than there.


SirAllap:
Merged — thank you. Second one of yours today, and this one reviewed even faster than the last.

The choice that carries it is reading Codex's quota off the rollout files instead of asking an endpoint for it. No credentials, no network, nothing that can rate-limit you for checking how rate-limited you are — and then `observedAt` and the refresh ping to pay for the staleness that buys, which is the part a shortcut would have skipped. The `resets_at` note is the detail I would have gone looking for: unix seconds read as milliseconds lands in 1970 and renders every reset as "now", and it says so in the file rather than in a commit message nobody re-reads.

Sorting the windows by length rather than by the `primary`/`secondary` keys is right too — those are positional and swap between plans, so the 5h meter would have rendered below the weekly one it sits inside on some accounts and not others.

I checked the surface rather than assuming it: no new egress, the Anthropic token path untouched apart from `CRED_PATH` becoming `credPath()` for test isolation, `refreshCodexUsage` spawning an argv array with no shell under `--sandbox read-only` with a timeout, and `/usage/codex/refresh` gated on `trustedCaller` rather than `localOrigin` — which is the stricter of the two and the right one for a route that runs a CLI. `/usage/providers` staying an open read is correctly argued in the comment: there is no path on disk in that payload.

Merged straight in — it was the only open PR already sitting on current `main`. 1770 pass in `server/`, 1391 in `web/`, both typechecks clean.

---

PR #467, opened 2026-08-05
fix(chat): a prompt queued behind a running turn is accepted, not broken

AlexBaldwin42:
## What does this PR do?

Fixes the tmux chat engine reporting an **error** for a prompt the CLI had accepted.

Typing into Claude Code while a turn is in flight does not fail and does not interrupt — the CLI **queues** the prompt and draws `Press up to edit queued messages` in its input box. `chatpane.ts` had no name for that state, so it read it as two different wrong ones:

- `inputBox()` takes the last `^❯` row as the box's contents, and that hint sits on the box's own row behind the box's own glyph. Read as content it is text that is neither empty nor the prompt we pasted — the exact signature of a picker having opened — so the turn reported **"This opened an interactive prompt in the chat's tmux pane"** when nothing had opened.
- On the *next* turn the hint was already on screen before the paste, so `waitPasted()` handed it back as though it were our text. Every Enter afterwards compared it against itself, looked swallowed, and pressed again — into a live pane, for `PASTE_TIMEOUT_MS`, until it gave up with **"the chat pane would not accept the prompt"**.

Both prompts had in fact been accepted, and both ran.

Seen on a machine whose network dropped mid-turn: the CLI sat retrying `Unable to connect to API (ENOTIMP) · Retrying in 8s · attempt 5/10`, two prompts sent from the chat queued behind it, and both came back as errors.

The change:

- **`BOX_HINTS`** — the input box's hint text reads as an empty box, not as typing. Covers the queued hint and the fresh-session `Try "…"`.
- **`QUEUED_RE` and a `queued` submit outcome** — a third state next to running and idle. Deliberately checked *below* the existing picker guard, so Enter still never reaches a `/model` or `/effort` menu, and *above* the box check, so we never press again and append a second copy of the prompt to the queue.
- **`waitDrained()`** — a queued turn waits for the queue rather than reporting anything. No deadline, since the turn ahead may legitimately run for minutes; bounded on pane liveness instead.
- **The transcript watermark is re-taken once the queue drains.** The turn we were queued behind wrote its answer *and* its end-of-turn line past our original watermark, so without this the fix would replay someone else's answer into this chat bubble and then close on their end marker.
- **A queued pane no longer reads as idle.** It says nothing and its box reads empty, which was indistinguishable from `/clear` and would have ended the turn ~8s in via `IDLE_PROBES_TO_END`.

Server-side only — no `shared/types.ts` change, no web change, no config or docs change.

## How was it tested?

Six new cases in `server/test/chat-pane.test.ts`, built on a `QUEUED` fixture transcribed from the pane where this was reported. They were verified to **fail against the current code with exactly the observed verdicts** (`diverted`, and `retry` until it times out to `stuck`), which is what pins them to the real bug rather than to the fix. The additions are purely additive — no existing test changed.

Everything CONTRIBUTING.md lists, run on this branch:

| | |
|---|---|
| `server`: `bunx tsc --noEmit` | pass |
| `server`: `bun test` | 1742 pass, 19 skip, 1 fail — see below |
| `web`: `bunx tsc --noEmit` | pass |
| `web`: `bun test` | 1357 pass, 0 fail |
| `web`: `bunx vite build` | pass |
| `bun scripts/smoke.ts` | pass — app mounted, no console errors |
| `bun scripts/perfbudget.ts` | pass — p99 3ms against a 120ms budget |

The one server failure is **pre-existing and environment-dependent, not from this change** — it reproduces identically on an unmodified checkout of `main`. `agent-routes.test.ts` → *"and refuses an agent this machine does not have"* asserts that Codex has neither a binary on PATH nor a config directory, and my machine happens to have both (`/usr/bin/codex`, `~/.codex`). It should pass on a runner without Codex installed. Happy to send a separate PR seaming that off the host if you'd like it.

### DIFF
--- a/server/src/chatpane.ts
+++ b/server/src/chatpane.ts
@@ -103,12 +103,36 @@ async function waitReady(name: string, deadline: number): Promise<boolean> {
  *  never accepted, and the engine reported failure on a successful turn. */
 const INPUT_ROWS = /^❯(.*)$/gm;
 
+/** Hint text the TUI draws *inside* the empty input box.
+ *
+ *  A hint sits on the box's own row, behind the box's own glyph, and is not one
+ *  character the user typed — but it is not whitespace either, so a box showing
+ *  one reads as "holding something" unless it is named here.
+ *
+ *  That distinction is load-bearing. When a turn is already running, a newly
+ *  submitted prompt is QUEUED, and the box then hints `Press up to edit queued
+ *  messages`. Read as content, that hint is text which is neither empty nor the
+ *  prompt we pasted, which is precisely the signature of "a picker opened" — so
+ *  a prompt the CLI had accepted was reported to the user as an interactive
+ *  prompt it could not draw. Worse on the next turn: the hint was already there
+ *  before the paste, so `waitPasted` returned it as though it were our text, and
+ *  every Enter afterwards compared the hint against itself and looked swallowed.
+ *  The turn pressed Enter into a live pane until it timed out and then reported
+ *  that the pane would not accept the prompt. Both prompts had in fact been
+ *  queued, and both ran. */
+const BOX_HINTS: RegExp[] = [
+  /^\s*Try "/,                                  // an empty box in a fresh session
+  /^\s*Press up to edit queued messages\s*$/,   // an empty box with prompts queued
+];
+
 /** What is typed in the live input box right now, or `null` if it is not on
  *  screen at all (the TUI redraws while a turn runs). Note the box renders its
- *  empty state with U+00A0, which `\s` covers. */
+ *  empty state with U+00A0, which `\s` covers, and its hinted states with the
+ *  strings above — both of which are empty as far as this engine is concerned. */
 export function inputBox(screen: string): string | null {
   let last: string | null = null;
   for (const m of screen.matchAll(INPUT_ROWS)) last = m[1];
+  if (last !== null && BOX_HINTS.some((re) => re.test(last!))) return "";
   return last;
 }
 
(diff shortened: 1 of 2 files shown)


SirAllap:
Merged — thank you, and welcome.

This is a hard bug to have found, because every symptom of it was a lie. Two prompts that had been accepted and that both eventually ran were reported to the user as errors, in two different wrong ways, and the pane looked fine the whole time. Reading the input box's hint as content is exactly the kind of thing that only shows up when you go and capture the real screen — which you did, `ENOTIMP · attempt 5/10` and all.

Three things made the review quick. The captured `QUEUED` fixture, so the tests assert against a frame that actually happened rather than one written to fit the fix. Pinning the ordering — picker before queue — with `__submitVerdict` rather than a comment, given Enter in a `/model` menu writes someone's real settings.json. And the note on the re-taken watermark: skipping the previous turn's output looks wrong until you say that the transcript is where the dashboard hydrates history from anyway, which you did.

`waitDrained` having no deadline is the right call. The turn ahead can legitimately run for minutes — that is *why* ours was queued — and bounding liveness on `paneAlive` instead of on a clock is what keeps a slow turn from being reported as a failure.

One narrow thing I noticed and did not change, for the record rather than as a request: `/^\s*Try "/` in `BOX_HINTS` is unanchored at the end, so a prompt that genuinely begins `Try "npm test" first` would read as an empty box and `waitPasted` would never see it land. The queued hint you came for is anchored both ends and has no such edge. I will tighten the fresh-session one separately.

Verified against current `main` — merges clean, `tsc` clean, 1775 pass in `server/`.

---

PR #514, opened 2026-08-10
feat(tmux): reopen the session you left, instead of asking you to attach it

SirAllap:
## What does this PR do?

Makes agentglass come back to the tmux session you were working in, instead of asking you to attach it yourself.

Reported as a daily annoyance with a one-line repro: close the app, open it again, and the session is not there — `tmux attach -t <name>` in a terminal **outside** agentglass is what brings it back. After every rebuild.

It is self-inflicted, and it takes two losses to happen.

**The app hid what it had just put down.** The panel's client is normally the only client on that server, so closing agentglass detaches the session — and `listPanes` skipped any server with no client at all. The one server guaranteed to be clientless a second after the app closes is the server the app was using.

**And it forgot where it was.** `lastTarget`, the only place the socket was ever recorded, is a module variable that dies with the process. Its own comment said "null until a terminal has been opened once this run, which is the honest answer". It was honest, and it was half the bug.

So the socket and the session name are now written next to the rest of our settings, and discovery lets through exactly one detached server: the one this app was demonstrably attached to. **The filter is not removed** — its comment names a real hazard, that this project's own suite leaves servers on scratch sockets and a resurrect config restores real sessions into them, so an unattached server can be a convincing duplicate. The new rule is "a server this app has used is one it may still see", not "show every detached server".

### Opening on it, rather than on a fresh shell

Remembering only made the session *visible*. The panel still started a bare shell, so the work was running and merely not on screen, and the way back was still to type the attach into that new shell. There are three cases and each was measured on tmux 3.7, because the obvious spellings are wrong:

- **Session alive** — `attach-session -t $id`. By id, not name: a name is prefix-matched, so a remembered `work` lands silently in `workbench`.
- **Server alive, that session gone** — `attach-session` with no `-t`, the most recently used one. Starting a server here would add a session beside the work.
- **No server at all, the cold boot** — `new-session -A -s <name>`. Bare `tmux` starts the server and attaches you to the session **it** just made, while a resurrect/continuum setup restores the real ones behind you; you end up in `0` watching none of them, which is exactly how this was reported. `start-server` cannot stand in either — a server with no sessions exits immediately. `-A` works because *creating* the session is what starts the server, which sources the user's configuration, which fires their restore — and resurrect puts missing windows into a session that already exists, so the one we just made is filled in rather than competed with.

Deliberately not `attachArgvFor`, the phone's attach: that one builds a grouped session so two screens can differ in size and marks `window-size` to hand back. The desk **is** the session.

### What counts as "where you work"

"The last session the panel attached to" is a different question, and answering the wrong one is visible: after a restore, tmux put the client on a one-window session on its way past and the memory recorded that. A cold start would then have brought back that session instead of the one the day was spent in. A session now has to **settle for 30 seconds** before it is written, and a phone's mirror never counts.

### A safety net for the user's resurrect saves

This one comes from losing a session for real while building the rest. resurrect keeps every save and a `last` symlink at the newest; continuum saves on a timer. A save fired while the server was being killed, wrote a file with **one pane** in it, and `last` moved to it — leaving a nine-pane save from ten minutes earlier on disk and unreachable by anything that restores. The same shape had already emptied `assistant-sessions.json` once before, so it is not a freak: a save on a dying or newborn server is poorer than the one it replaces, and resurrect has no notion of "poorer".

agentglass does **not** race that timer — two savers on one server is its own way to lose sessions. It keeps a copy where nothing else writes, and repoints `last` only when the live file is one pane against several, keeping what it replaced as `last.clobbered`. Their restore does the restoring; this only protects the file it reads. Panes rather than bytes, because `@resurrect-capture-pane-contents` makes size a function of scrollback, so an empty server's save can be the larger file.

Rules held throughout: their saves are read, never written and never deleted; the only write into their directory is that symlink; nothing is installed; and a machine with no resurrect at all sees none of this happen.

## How was it tested?

`tsc --noEmit` clean for `server/`.

**`server/test/tmux-memory.test.ts`** — 18 cases against **real tmux**, because every claim is about what tmux reports for a server nobody is attached to. Each test gets its own socket directory under a private `TMUX_TMPDIR`, and every server runs with `-f /dev/null`, so nothing here can find or read the developer's own tmux. The bug itself is pinned: a detached server, `list-clients` verified empty, is listed once remembered — and is **not** listed when it is not ours, when nothing is remembered, or when its server has since been killed. The attach builder is asserted to resolve by id with `work`/`workbench` standing side by side, to use `-A` on a cold socket, and to fall back to bare `tmux` when no name was ever recorded.

**`server/test/tmux-snapshot.test.ts`** — 15 cases: a one-pane save is never copied (it is the clobber's own shape), a smaller save somebody *meant* is left alone, `last` is repointed only on the real signature, what it replaced is kept, none of the user's files are destroyed, and the pointer is never absent mid-repair.

52 tests pass across these two plus the recipes suite when both features are built together.

Driven on the machine that reported it, through a desktop build, including a real `tmux kill-server` and cold start.

**One thing worth saying plainly**: while building the snapshot module, its own test wrote into the developer's real `~/.tmux/resurrect` and destroyed a save. `os.homedir()` does not follow `$HOME` on Bun, so a suite that redirected `HOME` was not isolated at all, and a fixture helper writing with `fs` bypassed every guard the module had. Both are fixed here — the module reads `$HOME` first, and the helper **throws** rather than writing outside the scratch directory — and the episode is written into the comments so the next person does not rediscover it the same way.

### DIFF
--- a/server/src/index.ts
+++ b/server/src/index.ts
@@ -102,6 +102,7 @@ import { generateWalkthrough, WALKTHROUGH_ENABLED } from "./walkthrough.ts";
 import { ptyOpen, ptyMessage, ptyClose, projectCommands, shutdownTerminals, lastTmuxTarget, sessionTitle, TERMINAL_ENABLED, PTY_BACKEND, type PtyWsData } from "./terminal.ts";
 import { mintAgentTicket } from "./agentticket.ts";
 import { listPanes, focusPaneAnywhere, activePane, sweepPinnedWindows, pinnedSockets } from "./tmuxctl.ts";
+import { repairLast, snapshot } from "./tmuxsnapshot.ts";
 import { withAgentSessions } from "./paneloc.ts";
 import { notePaneFromHook, paneDirs, paneAgentNote } from "./panewt.ts";
 import { chatSend, activeTurns, CHAT_ENABLED, CHAT_BYPASS_ALLOWED, CHAT_ENGINE_DEFAULT } from "./chat.ts";
@@ -3433,6 +3434,33 @@ for (const sig of ["SIGINT", "SIGTERM"] as const) {
  */
 const unpinned = sweepPinnedWindows(pinnedSockets());
 if (unpinned) console.log(`🪟 tmux: ${unpinned} window(s) released — a previous run was killed before it could put window-size back`);
+
+/*
+ * Keep a copy of the user's last good resurrect save, and put `last` back if a
+ * dying server's save has landed on top of it.
+ *
+ * Measured on a real machine, and it cost a working session: a save fired while
+ * tmux was being killed, wrote a file with one pane in it, and `last` moved to
+ * that — leaving a nine-pane save from ten minutes earlier on disk and
+ * unreachable by anything that restores. resurrect has no notion of "poorer",
+ * and racing continuum's timer would be two savers on one server, which is its
+ * own way to lose sessions. So: copy, and repair the pointer THEIR restore
+ * reads. See tmuxsnapshot.ts for what it will and will not touch.
+ *
+ * At boot and then every few minutes. A tick that finds the same save does
+ * nothing at all, which is nearly every tick.
+ */
+const SNAPSHOT_TICK_MS = 4 * 60 * 1000;
+function guardResurrect(): void {
+  const put = repairLast();
+  if (put) {
+    console.log(`💾 tmux: restored resurrect's "last" to ${put.restored} — a save from a dying server had replaced it (the poor one is kept as last.clobbered)`);
+  }
+  const took = snapshot();
+  if (took) console.log(`💾 tmux: kept a copy of ${took.split("/").pop()}`);
+}
+guardResurrect();
+setInterval(guardResurrect, SNAPSHOT_TICK_MS).unref?.();
 
 // Parent-death watchdog: a server must never outlive whoever launched it.
 //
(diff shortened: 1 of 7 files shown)


SirAllap:
Two commits added after opening this, both from asking what happens on a machine that is not the one it was built on.

**No tmux at all.** A missing binary and an unstarted server answer identically — both make the tmux call return `null` — so the cold-boot branch would have handed the panel `tmux new-session …` to run where there is no tmux, and the terminal would open on `command not found` instead of on a shell. It needs tmux to have been present when the memory was written, so it is narrow, and it is still a broken terminal. Now refused explicitly, with a test.

The PATH is passed to `Bun.which` rather than left implicit: measured on 1.3.9, `Bun.which()` snapshots the PATH and mutating `process.env.PATH` does not change its answer, while the `{ PATH }` option does. The bare call was therefore answering about the environment the process started in, and could not be asked about a machine without tmux at all — which is also why the first version of that test passed while proving nothing.

**No resurrect (but tmux present)** needed no change, and was measured rather than assumed: `readLast`, `snapshot`, `clobbered` and `repairLast` all answer null and do nothing at all. The session resume itself does not depend on resurrect — after a reboot you get the session back by name, empty, because there is nothing on the machine that restores its windows. That is the whole of what can be done without becoming a second saver.

Also scrubbed a real session name out of the comments; the behaviour they describe is about prefix matching, not about that name.

PR #529, opened 2026-08-16
ci: use locked installs and local web build

### LINKED ISSUE #470: CI resolves its own toolchain from npm on every run
Four unpinned inputs in `.github/workflows/ci.yml`, in rough order of exposure. `bun.lock` and `Cargo.lock` are committed, so a local install is reproducible — the gap is CI, which is the thing that runs on every contributor's branch.

**1. `bunx` resolves and executes latest from npm.**
```yaml
run: cd web && bunx tsc --noEmit
run: cd server && bunx tsc --noEmit
run: cd web && bunx vite build
```
`server/` has no local typescript at all, so every CI run downloads a fresh one and runs it. Whatever npm serves that minute is what typechecks and builds this repo.

**2. `bun install` runs without `--frozen-lockfile`.**
```yaml
run: bun install
```
So `bun.lock` can drift in CI without anything failing — which is the one place drift should fail loudly, because it is the run that decides whether a PR is green.

**3. The toolchain itself is a moving target.**
```yaml
bun-version: latest
```
A Bun release can turn CI red on a branch that changed nothing, and it has no bearing on what anyone runs locally.

**4. Third-party actions are on moving tags.**
`actions/checkout@v7`, `oven-sh/setup-bun@v2`, `browser-actions/setup-chrome@v2`. A tag is mutable; a commit SHA is not.

### What to do

In order of value for effort: add `--frozen-lockfile`; add `typescript` and `vite` as dev dependencies and call them from `node_modules` instead of `bunx`; pin `bun-version` to an exact release; pin the actions to SHAs with a comment naming the version.

The `bunx tsc` one has bitten already — it is what #458's author found crashing locally, and it is the same resolve-at-run-time behaviour either way.

Raised by @dakristia in the security review on their fork (#458, closed). Filed here so it does not disappear with that branch.

ayushcodes13:
## What does this PR do?

Closes #470.

This makes the CI workflow use the repository's locked dependency graph more consistently:

- changes workspace installs to `bun install --frozen-lockfile`
- replaces the remaining `cd web && bunx vite build` with `cd web && bun run build`, so CI uses the repo-local Vite dependency declared in `web/package.json`

I intentionally left exact Bun/action SHA pinning out of this PR to keep the change small and reviewable.

## How was it tested?

Passed locally:

```bash
bun install --frozen-lockfile
cd web && bun run typecheck
cd web && bun run build
cd server && bun run typecheck
git diff --check
```

Additional local suite notes:

```bash
cd server && bun test --timeout 20000
```

This ran but failed in this macOS shell on environment-dependent tests, primarily missing `tmux`, process/machine introspection differences, and long-running desktop install/process cases. Many unrelated server tests passed before those failures.

```bash
cd web && bun test
```

This ran 2716 web tests: 2714 passed, 2 failed in `workspace chords` key-label expectations on this platform. The CI-targeted typecheck and production build both passed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         with:
           bun-version: latest
       - name: Install
-        run: bun install
+        run: bun install --frozen-lockfile
       # The mark lives in seven places — a standalone SVG, the favicon, a React
       # component, the boot splash and three inlines in the landing page. They
       # drifted once and a fix shipped to six of them while the seventh kept
@@ -52,7 +52,7 @@ jobs:
       - name: Tests (web)
         run: cd web && bun test
       - name: Build web
-        run: cd web && bunx vite build
+        run: cd web && bun run build
       # Pinned rather than the runner's preinstalled Chrome: the ubuntu-latest
       # image has dropped and re-added browsers before, and this also keeps the
       # step working on container or self-hosted runners that ship none.
@@ -149,7 +149,7 @@ jobs:
       # the only job that runs the server outside `build`.
       - name: Install (workspace — the server those two suites boot)
         if: steps.phone.outputs.there == 'yes'
-        run: bun install
+        run: bun install --frozen-lockfile
       # Both configs: the app against React Native, the suites against
       # bun-types. `types` replaces rather than adds, so one file cannot hold
       # both — see mobile/tsconfig.test.json.


SirAllap:
Thank you for this, @ayushcodes13 — genuinely useful cleanup, and the restraint of leaving SHA pinning out to keep the diff reviewable was the right call.

Verified before merging:

- `bun install --frozen-lockfile` resolves against the committed `bun.lock` with no drift, so this won't red CI on the current tree.
- `cd web && bun run build` builds clean, and `web/package.json` declares `build: vite build` — so CI now uses the repo-local Vite instead of whatever `bunx` resolves.
- This also brings `ci.yml` in line with `desktop-binaries.yml`, which already installs frozen and builds through the script. That drift was exactly the thing worth closing.

One leftover for whoever picks up #470 next: `.github/workflows/pages.yml` still runs a bare `bun install`. Out of scope here.

Merging. Thanks again.

---

PR #531, opened 2026-08-17
fix(build): darwin links the way darwin links

SirAllap:
## What does this PR do?

The macOS jobs get past ncurses now and die in tmux `configure`, one way per arch. Both messages are Linux assumptions landing on a platform that does not share them.

- **`-static`.** There is no static libSystem on macOS and the linker will not pretend there is, which reaches configure as the flat "C compiler cannot create executables". What the app actually needs is to depend on no tmux, libevent or ncurses the user may not have — and those are `.a` archives built in this script, so they link in either way. libSystem stays dynamic, as it does for every binary on that platform.
- **`-s`.** A GNU ld flag; obsolete on Apple ld. `-Wl,-x` is the local-symbol strip that means the same thing there.
- **`--host=arm64-apple-darwin`.** tmux 3.5a ships a `config.sub` that does not know that spelling and stops the build with "config.sub … failed". A native build does not need to be told what it is, and `-arch` is what decides the slice — which is the one thing the x64 job has to say out loud on an arm64 runner.

## How was it tested?

Built end to end on linux after the refactor: `out/tmux-bun-linux-x64`, `statically linked, stripped`, and `tmux 3.5a` from the binary. That is the path these three variables could have broken, and it is unchanged. macOS is what the release run answers.

### DIFF
--- a/scripts/build-tmux-static.sh
+++ b/scripts/build-tmux-static.sh
@@ -75,6 +75,27 @@ case "$TARGET" in
   *darwin*) NCURSES_TERMLIB="";                     TINFO_LDLIB=""        ;;
 esac
 
+# How far "static" goes, and how the target is named — both differ on darwin,
+# and both were assumed to be the Linux answer.
+#
+#   -static: there is no static libSystem on macOS and the linker refuses to
+#     pretend otherwise, which reaches `configure` as the flat "C compiler
+#     cannot create executables". What the app actually needs is no dependency
+#     on a tmux, libevent or ncurses the user may not have — and those are
+#     already `.a` archives built right here, so they link in either way.
+#     libSystem stays dynamic, as it does for every binary on that platform.
+#   -s: a linker flag on GNU ld, obsolete on Apple's. `-Wl,-x` is the local
+#     symbol strip that means the same thing here.
+#   --host: tmux 3.5a's `config.sub` does not know `arm64-apple-darwin` and
+#     stops the build with "config.sub … failed". A native build does not need
+#     to be told what it is; `-arch` is what decides the slice, and it is the
+#     one thing the x64 job needs to say out loud on an arm64 runner.
+case "$TARGET" in
+  *linux*)          LDMODE="-static -s"; ARCHFLAG="";              HOSTFLAG="--host=$TRIPLE" ;;
+  bun-darwin-arm64) LDMODE="-Wl,-x";     ARCHFLAG="-arch arm64";   HOSTFLAG=""               ;;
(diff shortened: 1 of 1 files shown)


SirAllap:
Reviewed this against the current tree. The three diagnoses hold — no static libSystem, `-s` being GNU-only, and tmux 3.5a's `config.sub` not knowing the `arm64-apple-darwin` spelling — and the diff still applies cleanly to `main`. One real gap and one risk, both confined to the `bun-darwin-x64` job, which `desktop-binaries.yml:31-33` cross-builds **on a `macos-14` arm64 runner**:

**1. `build_libevent` never gets `$ARCHFLAG`.** `build_ncurses` and `build_tmux` both received it; libevent didn't. On the x64 job that produces an arm64 `libevent_core.a` alongside x86_64 ncurses and tmux objects, and the tmux link fails on mismatched slices. Fix is mechanical:

```sh
CC="$CC" CFLAGS="-O2 $ARCHFLAG" LDFLAGS="$ARCHFLAG" ./configure $HOSTFLAG --prefix="$WORK/prefix" ...
```

**2. Dropping `--host` turns the x64 job into a cross-build that claims to be native.** Correct for `bun-darwin-arm64`, where the runner really is the target. On the x64 job, autoconf will compile its test executables `-arch x86_64` and then try to *run* them on arm64 — that's "cannot run C compiled programs" without Rosetta, or silently wrong feature detection with it. The `config.sub` complaint is specific to the `arm64-` spelling; `x86_64-apple-darwin` is a triple `config.sub` does know, so keeping `HOSTFLAG` for the x64 target and clearing it only for arm64 gets both properties:

```sh
bun-darwin-arm64) LDMODE="-Wl,-x"; ARCHFLAG="-arch arm64";  HOSTFLAG=""                          ;;
bun-darwin-x64)   LDMODE="-Wl,-x"; ARCHFLAG="-arch x86_64"; HOSTFLAG="--host=x86_64-apple-darwin" ;;
```

**Adjacent, pre-existing — this PR will hit it next.** `desktop-binaries.yml:108` runs `./out/tmux-${{ matrix.bunTarget }} -V` and then the `new-session` smoke test, both on the build runner. For `bun-darwin-x64` that's an x86_64 binary executed on the arm64 runner. Once the two items above are fixed, that's the next thing the release run will die on — the smoke test needs to be skipped when the target arch isn't the runner's.

Holding this one rather than merging: a release run is expensive to burn, and the libevent flag is a one-line fix. Everything else here is right, including keeping the `.a` archives linked in while libSystem stays dynamic.

---

PR #533, opened 2026-08-19
feat(web): pin custom terminal commands

### LINKED ISSUE #18: Terminal: pin a command you type, not only one the project offers
The terminal header has a QUICK row of one-click commands — but it's hardcoded (`git status`, `git log --oneline -15`, `git diff --stat`, `git branch -vv` in `web/src/components/TerminalPanel.tsx`). Everyone's muscle memory is different: if your fingers say `c` for clear and `..` for `cd ..`, you want *your* chips there, not mine.

(Worth noting: real shell aliases already work — the PTY spawns your login shell with `-il`, so anything in your `.bashrc`/`.zshrc` behaves exactly as in a normal terminal. This issue is about the *panel's* clickable chips, which are useful from the mouse and discoverable in a way rc-file aliases aren't.)

## Proposal

- A small **edit mode for the QUICK row**: add / remove / reorder chips, each with a label and the exact command it types (commands are typed into the live shell, same as today — no new execution path, no server changes needed for v1)
- Persist in `localStorage` first (matches how the panel already remembers the repo); a later iteration could move them to the config file so they sync across browsers
- Defaults stay as the current four for new users
- Nice-to-have: per-repo chips (a project's own favorite commands) on top of the global set — the ⚙ commands menu already proves the per-repo pattern

## Acceptance criteria

- I can add a chip `c` → `clear` and `..` → `cd ..`, and they survive a reload
- Deleting/reordering works, defaults restorable
- Chips still respect the existing disabled states (demo mode, terminal disabled)

Entry points: `QUICK` const and the chip row in `web/src/components/TerminalPanel.tsx`; `runInShell` is the only execution primitive needed.

Solaris-star:
## What does this PR do?

Fixes #18 by adding a free-text entry to the terminal Commands dropdown. A custom pin stores a separate chip label and command, so `c` can run `clear`; pins remain per-repository, retain the existing five-pin cap, and survive reloads through the existing localStorage store. Existing command and recipe pins remain backward-compatible.

## How was it tested?

- `bun run typecheck` in `web/` — passed.
- `bun test test/command-bar-pins.test.ts` — 3 passed.
- `bun run build` in `web/` — passed.
- `bun run perfbudget.ts` — passed (p99 6ms under the 120ms budget).
- `bun test` in `web/` — 2717 passed, 2 pre-existing failures in `test/chords.test.ts` (`ctrl+j`/mac platform expectations), unrelated to this change.
- `bun test` in `server/` — 2764 passed, 108 skipped, 108 failures and 10 errors; the failures are existing process/install test timeouts, unrelated to the web-only change.
- `bun scripts/smoke.ts` — app mounted, but the existing workspace-chord assertion reports duplicate `⌘1`/`⌘2`; unrelated to this change.

The commit is SSH-signed with the contributor's GitHub signing key.

### DIFF
--- a/web/src/components/CommandBar.tsx
+++ b/web/src/components/CommandBar.tsx
@@ -50,6 +50,46 @@ export const GIT_COMMANDS: ProjectCommand[] = [
 export const RECIPE = "recipe:";
 const MAX_PINS = 5;
 const PIN_KEY = "agentglass.commandPins";
+const CUSTOM_PIN = "custom:";
+
+/** A command the person typed themselves, rather than one discovered in the repo. */
+export type CustomPin = { label: string; cmd: string };
+
+/**
+ * Keep the old string[] storage format while giving custom pins their own
+ * display label. Existing command and recipe pins therefore survive the
+ * upgrade untouched, and a label containing spaces or punctuation is safe in
+ * localStorage too.
+ */
+export function encodeCustomPin(label: string, cmd: string): string {
+  const name = label.trim();
+  const command = cmd.trim();
+  if (!name || !command) return "";
+  return `${CUSTOM_PIN}${encodeURIComponent(JSON.stringify({ label: name, cmd: command }))}`;
+}
+
+export function decodeCustomPin(value: string): CustomPin | null {
+  if (!value.startsWith(CUSTOM_PIN)) return null;
+  try {
+    const parsed = JSON.parse(decodeURIComponent(value.slice(CUSTOM_PIN.length))) as Partial<CustomPin>;
+    if (typeof parsed.label !== "string" || typeof parsed.cmd !== "string") return null;
+    const label = parsed.label.trim();
+    const cmd = parsed.cmd.trim();
+    return label && cmd ? { label, cmd } : null;
+  } catch {
+    return null;
+  }
+}
+
+export function addCustomPin(root: string, label: string, cmd: string): boolean {
+  const encoded = encodeCustomPin(label, cmd);
+  if (!encoded || !root) return false;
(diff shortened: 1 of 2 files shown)


SirAllap:
Thank you, @Solaris-star — this is a careful piece of work and it closes #18 the way it should have been closed.

The part I appreciated most: keeping the `string[]` storage shape and giving custom pins a `custom:` prefix, rather than migrating the store. Existing command and recipe pins survive untouched, the five-pin cap and the per-repo keying still hold, and `decodeCustomPin` returning `null` on anything that isn't yours means an ordinary pinned command can never be misread as a custom one. The round-trip test for punctuation in labels and commands is the right test to have written.

Verified on a local merge with `main`:

- `bun run typecheck` in `web/` — clean (both configs).
- `bun test` in `web/` — **2719 pass, 0 fail**. The two `chords.test.ts` failures you saw are macOS platform-label expectations; they pass on Linux, so nothing to chase.
- `bun run build` — clean.

Three small things, none of them blocking, noted here so they aren't lost:

1. `addCustomPin` returns `false` both when the pin cap is hit and when `root` is empty, so `saveCustom` reports "You already have 5 pinned commands" in a case where that isn't why it failed.
2. `{placeholder}` prompting won't fire for a custom pin — `encodeURIComponent` escapes the braces, so the check in the chip-click path never sees them. Defensible (you typed the command yourself), just worth a comment in the code.
3. `decodeCustomPin(cmd)` is called three times per chip in the title/label expressions.

Merging. Thanks for the thorough test notes in the description too — knowing which failures were pre-existing saved a round trip.

---

PR #548, opened 2026-09-04
feat(mobile): Pane — the phone gets a design system of its own

SirAllap:
## What does this PR do?

The phone stops wearing github-dark and gets a palette, a corner ladder and a type scale of its own — **Pane**. The desk is untouched: `paletteFor` and `accentFor` now take the base palette as an argument instead of reading the module constant, so one product changing its skin is a parameter rather than a fork.

**Colour.** PANE is not github-dark tuned. Its ground is darker (`#0a0c10`), its hairlines are quieter and its raised surfaces come *up* instead — a desk separates rows with borders because it has the pixels, a phone at 393 points separates them with surface. The mid inks go up, because a phone is read outdoors. The light half is warm rather than a negative of the dark one: a cool near-white reads as a screen left on, a warm one reads as paper. Teal is the shipped accent and lives in `ACCENTS` beside the other six, so the desk gains the row too — deliberately, since the rule that file already argues is that an accent named the same in both products *is* the same hex.

**Corners.** 8 / 10 / 16, plus a separate `pill` at 22. `sm` and `md` sit two points apart on purpose: everything you press is nearly a rectangle, and there is exactly one round thing per screen. The sheet's top edge stops being `RADIUS.lg + 6` and becomes that pill.

**Type.** Title 16 → 17, and Now's opening line gets a size of its own at 26. That line is the only Display in the app because it states a count — "3 things want you" — and is what the screen is for; a pull request's own title stays at 20, since 26pt of somebody else's prose is four lines before the screen has said anything. The eyebrow keeps 11 rather than dropping to the design's 10.5: this is read outdoors, in one hand, by somebody deciding whether a check failed, and what makes an eyebrow read as one is the tracking and the caps, which `Label` has carried all along.

**No font files.** Pane wants one family for prose and code, which means bundling IBM Plex through `expo-font`, a load-before-render step and ~400 KB of APK for a difference nobody on a phone reads as a family choice. The ladder, the tracking and the mono/prose split all work on the system face, which is where the idea actually lives.

Three faults were found while wiring it, each of which would have shipped:

- `setLook` composed the live repaint on `BASE`. The app would have looked like Pane until the first time anybody opened Settings, and then snapped back to github-dark for good.
- `accentFor`'s unknown-id fallback read the module `BASE` instead of its own argument, so an unreadable preference file put the desk's blue on a Pane ground.
- `info` was set to the teal, matching the accent. Wrong twice: in the UI it collapses "informational" and "press me" into one hex, and in the terminal `deriveAnsi` picks ANSI blue by *hue* — teal is 171°, so blue fell through to `primary`, and on the neutral accent `primary` **is** the body text. Two of the sixteen slots held the same colour and an `ls` colour vanished.

Two hardcoded colours went with it. The diff's added and removed rows carried github's green and red written out as `rgba()` long after the phone stopped wearing them — with the red a whole hue from the `C.error` printed on the same row — and now derive from the live palette via `tint()`. The two scrims (`rgba(1,4,9,.55)` on one screen, `rgba(0,0,0,.55)` on another) became one `SCRIM`.

No server routes, no permission or scope changes, no `shared/types.ts` contract change.

## How was it tested?

- `npx tsc --noEmit` in `mobile/` — clean.
- `bun test` in `mobile/` — 579 pass, 0 fail, 40 skip across 44 files.
- `bun test` in `web/` — 2716 pass, 0 fail across 210 files, confirming the desk is unchanged by the shared/palettes.ts edits.
- The `info` fault above was caught by `mobile/test/term-ansi.test.ts` once its palettes were composed on what the app actually paints rather than on the module default — that file now builds every palette through a `phone()` helper, so its WCAG floors are measured against the real ground.
- New coverage: `tint()` is checked for two-digit padding (a one-digit alpha makes a seven-character colour that RN silently draws as transparent), clamping at both ends, and following the live palette; `accentFor`'s fallback is checked against a caller-supplied base.

Not exercised on a device — no phone attached to this environment. The palette, radius and type changes are declarative and covered by the suites above, but a visual pass on a real screen is worth doing before merge.

## CI

`build` is red on the **Tests (server)** step, and it is not this branch's — see the comment below: `main`'s own last green run, re-run today unchanged, fails identically.

### DIFF
--- a/mobile/app/(tabs)/now.tsx
+++ b/mobile/app/(tabs)/now.tsx
@@ -148,7 +148,7 @@ export default function NowScreen(): React.ReactNode {
       }
       ListHeaderComponent={
         <View style={{ gap: SPACE.xs, paddingBottom: SPACE.sm }}>
-          <Text style={{ color: C.text, fontSize: T.head, fontWeight: "700" }}>
+          <Text style={{ color: C.text, fontSize: T.display, fontWeight: "700", lineHeight: 31 }}>
             {waiting.length === 0
               ? "Nothing is waiting on you"
               : `${waiting.length} ${waiting.length === 1 ? "thing wants" : "things want"} you`}
--- a/mobile/app/pr/diff.tsx
+++ b/mobile/app/pr/diff.tsx
@@ -46,7 +46,7 @@ import {
 import { gapLabel, gapsIn, nextSlice, type Gap } from "../../src/model/expand.ts";
 import { draftCount, takeDraft, type LineNote } from "../../src/model/reviewDraft.ts";
 import { Btn, Card, Label, Note, Sheet, SheetRow, TAP } from "../../src/ui.tsx";
-import { C, MONO, RADIUS, SPACE, T } from "../../src/theme.ts";
+import { C, MONO, RADIUS, SPACE, T, tint } from "../../src/theme.ts";
 
 /** The two backgrounds a changed line takes.
  *
@@ -55,8 +55,8 @@ import { C, MONO, RADIUS, SPACE, T } from "../../src/theme.ts";
  *  marker column carries the sign as well, so the distinction survives for
  *  anybody who cannot rely on the tint. */
 function lineFace(kind: DiffLine["kind"]): { bg: string; mark: string; ink: string } {
-  if (kind === "add") return { bg: "rgba(63,185,80,0.14)", mark: "+", ink: C.success };
-  if (kind === "del") return { bg: "rgba(248,81,73,0.14)", mark: "−", ink: C.error };
+  if (kind === "add") return { bg: tint(C.success, 0.14), mark: "+", ink: C.success };
+  if (kind === "del") return { bg: tint(C.error, 0.14), mark: "−", ink: C.error };
   return { bg: "transparent", mark: " ", ink: C.text4 };
 }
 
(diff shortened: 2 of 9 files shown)


SirAllap:
## `build` is red, and it is not this PR's

`build` fails on the **Tests (server)** step with 10 failures, all in `server/test/usage-across-the-seam.test.ts` — the usage rollup's retention-boundary suite. Every other check is green, `mobile` included.

**The decisive evidence:** `main`'s own last green run — commit `e697196`, run `33201139674`, green on 2026-08-28 — was re-run today, unchanged, and now fails with the identical result: `2964 pass / 19 skip / 10 fail`, the same ten test names. The same commit that passed a week ago fails now. That is the base branch being red, not this branch breaking it.

The workflow's own comment above the `setup-bun` pin describes this exact failure mode happening once before ("re-running the CI of a commit from the 17th, untouched, fails today and passed then"), which is why the version is pinned at all.

**Three further checks, none of which implicate this PR:**

- The diff touches `mobile/**` and `shared/palettes.ts` only. The server never imports `shared/palettes.ts` — the single match in `server/` is the word "palettes" inside a prose comment in `themesync.ts`.
- Full server suite run on this branch and on `origin/main`, same machine, same minute: **zero** failures unique to this branch.
- Not reproducible locally at all. Runs clean in isolation and in the full suite, under bun 1.3.11 and under CI's pinned 1.3.14, in UTC (this box and the runner are both UTC, so it is not a timezone split). The failure appears only on the runner.

**No root cause yet, and no fix to port.** The suite passes everywhere except the GitHub runner, so the cause is in the interaction between that environment and this suite rather than in the assertions themselves — but I could not reproduce it, so I am not going to guess at a patch in a PR about colours. It needs its own change against `main`, with a reproduction first.

I have not touched, skipped, or weakened those tests to make this branch green, and I won't.

**What this means for merging:** this PR does not touch the server, and `main` is red on the same step today regardless of it. Whether that is enough to merge past a red check is your call, not mine.

---

PR #549, opened 2026-09-04
fix(mobile): the terminal's bottom bar stops being five kinds of furniture

SirAllap:
## What does this PR do?

Stacked on #548 — it uses `RADIUS.pill`, which that PR introduces. Merge #548 first and this retargets to `main` on its own.

The terminal's bottom bar had five different kinds of thing on one row: a pinned island holding `fit` and `80c`, a `line`/`keys` toggle, a text field, and a set of icon buttons, each with its own fill, border and radius. Compared side by side with Orca's, the difference is not that ours has more features — it is that ours looks like five components that happened to end up adjacent.

- **`80c` is gone.** It sets the column count, which already lives in the terminal's settings. A control that duplicates a setting is a second place for the same value to be wrong.
- **`line` is gone.** It toggled between two input modes, and the mode it selected is now said by the field itself: in key mode the field's border takes the accent. The switch existed to announce a state that the state can announce on its own.
- **`fit` is gone** from the bar. It was unlabelled in effect — nothing on the screen said what it fit or to what.
- Both survivors moved into the `···` sheet under a **This pane** heading, where they read as settings for this pane rather than as controls competing with the place you type.

What is left is one row of keys and one capsule. The field and its three icons are a single pill: one border, one background, one radius, with the icons carrying no fill at all except send (the accent) and the mic while it is listening (the error colour). That is the palette's rule applied here — colour means state, and an idle control is not a state.

The field's height goes to `TAP` (44) rather than the 40 it was borrowing from an allowance argued for keys. It costs nothing, since the icons beside it are already 44.

Bar capacity goes from 7 keys to 11 in the same width, because the pinned island is no longer taking a fifth of the row.

## How was it tested?

- `npx tsc --noEmit` in `mobile/` — clean.
- `bun test` in `mobile/` — 579 pass, 0 fail, 40 skip.
- `mobile/test/tap-floor.test.ts` passes without a new allowance entry: the field is 44 and the icons are 44, so nothing here needed one. An earlier draft put a checklist row at `minHeight: 26` and this suite caught it; the height was removed rather than allowed, since that row is read-only and not a target.

Not exercised on a device — no phone attached to this environment. The one thing worth a visual check before merge is the capsule against the key row at 393 points, since the keys and the pill are the two things now sharing that space.

### DIFF
--- a/mobile/app/(tabs)/terminal.tsx
+++ b/mobile/app/(tabs)/terminal.tsx
@@ -1718,7 +1718,10 @@ export default function TerminalScreen(): React.ReactNode {
             // takes the accent, because the field behaving differently is the
             // thing that switch used to have to say out loud.
             borderColor: raw ? C.primary : C.border,
-            borderRadius: RADIUS.lg, paddingLeft: SPACE.md, paddingRight: 3,
+            // The capsule, and the only one on this screen. Pane allows exactly
+            // one round thing per screen against everything else being nearly
+            // rectangular, and on the terminal this is it: the place you type.
+            borderRadius: RADIUS.pill, paddingLeft: SPACE.md, paddingRight: 3,
             paddingVertical: 3,
           }}>
           <TextInput
--- a/mobile/app/(tabs)/terminal.tsx
+++ b/mobile/app/(tabs)/terminal.tsx
@@ -86,7 +86,7 @@ interface AgentOffer {
 }
 import { bestSession, readStrip, sessionsOf, type Tab } from "../../src/terminal/tabs.ts";
 import type { PanesResponse } from "../../../shared/types.ts";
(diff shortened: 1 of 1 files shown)


SirAllap:
## `build` is red here for the same reason as #548

Same failing step (**Tests (server)**), same 10 tests in `server/test/usage-across-the-seam.test.ts`. This branch is stacked on #548 and inherits it; neither branch touches the server.

The evidence is written up once, on #548: https://github.com/SirAllap/agentglass/pull/548#issuecomment-5547260460 — the short version is that `main`'s own last green run (commit `e697196`, green 2026-08-28) was re-run today unchanged and fails identically, so the base branch is red on this step regardless of either PR.

Not re-running the job here: it is the same failure on the same base, and a second confirmation of an already-confirmed result buys nothing. No tests were touched, skipped, or weakened on either branch.

---

PR #561, opened 2026-09-06
fix(security): gate configured egress destinations

### LINKED ISSUE #468: Two egress paths leave this machine with no check on where they go
Both verified against `main` today. Neither is a bug in the sense of misbehaving — they do exactly what they were written to do. The gap is that every *other* outbound path in this app validates its destination and these two do not, so an environment variable is enough to redirect them.

Raised originally by @dakristia in the security review on their fork (#458, closed). Filing here so the findings outlive that branch — the code fixes from that review shipped in #465, these are the deferred half.

### 1. `AGENTGLASS_WEBHOOK` is POSTed to unvalidated — `server/src/alerts.ts:19`

```ts
const WEBHOOK = process.env.AGENTGLASS_WEBHOOK;
```

No local-only check, no allowlist, nothing logged about where it points. What goes out is short human-readable alert text — truncated command summaries, error strings, the agent label — not transcripts or tokens, so the blast radius is small. But it is the only egress in the codebase with no destination check at all, which sits oddly beside the hooks, where the same class of variable is hard-refused unless `AGENTGLASS_ALLOW_REMOTE=1`, and where #465 has just tightened that refusal.

Smallest useful fix: log the destination host once at boot, so an operator can never be sending alerts somewhere without knowing. Fuller fix: the `AGENTGLASS_ALLOW_REMOTE` shape the hooks already use.

### 2. The walkthrough's base URL comes from the environment — `server/src/walkthrough.ts:212`

```ts
const client = new Anthropic();
```

The SDK reads `ANTHROPIC_BASE_URL`, so the compressed diff hunks this sends — actual repository code — are redirectable by an env var alone, with no opt-in and nothing on screen to say so. This is the one feature that ships repo *content* to a model, which makes it the one where the destination matters most.

Fix: refuse a non-Anthropic base URL unless an explicit opt-in is set, same shape as `AGENTGLASS_ALLOW_REMOTE`.

### 3. And it has no kill switch — `server/src/walkthrough.ts:34`

```ts
export const WALKTHROUGH_ENABLED = !!claudeBin() || hasApiKey();
```

Derived from "is `claude` on PATH, or is a key set", so there is no way to turn it off. Every other capability has one: `AGENTGLASS_CHAT_DISABLED`, `AGENTGLASS_GIT_WRITE_DISABLED`, `AGENTGLASS_DOCKER_WRITE_DISABLED`, `AGENTGLASS_SCAN_DISABLED`, `AGENTGLASS_CODEX_DISABLED`, `AGENTGLASS_ANTIGRAVITY_DISABLED`. An `AGENTGLASS_WALKTHROUGH_DISABLED=1` would make this consistent, and matters more here than for most of them given what the feature sends.

be-student:
# Summary

- validate environment-configured outbound URLs at one shared boundary
- require `AGENTGLASS_ALLOW_REMOTE=1` for remote webhooks and custom remote Anthropic base URLs
- log the webhook host without exposing its credential-bearing path
- add `AGENTGLASS_WALKTHROUGH_DISABLED=1`

Closes #468.

# Validation

- `cd server && bun run typecheck`
- `cd server && NODE_ENV=test bun test test/egress-destinations.test.ts test/pr-nudge.test.ts test/alerts-sink.test.ts test/walkthrough-egress.test.ts` (66 pass)
- `bunx oxlint server/src/egress.ts server/src/alerts.ts server/src/prnudge.ts server/src/walkthrough.ts server/test/egress-destinations.test.ts`

The repository-wide server suite was also attempted. It was stopped after several minutes because unrelated environment-dependent tests were already failing on macOS path canonicalization, unavailable machine metrics, tmux fixtures, and a `Bun.serve({ port: 0 })` fixture.


SirAllap:
Thank you for this, @be-student — and welcome. #468 had been sitting open since the audit in #458, and you took the whole of it: the webhook, the base URL and the missing kill switch, behind one boundary instead of three ad-hoc checks.

The part I appreciated most: `outboundDestination()` compares `AGENTGLASS_ALLOW_REMOTE` against the literal `"1"`, the same shape `hooks/send_event.py` settled on and for the same reason — `=0`, `=false`, `=off` must never read as "on". And the refusal names `parsed.host` and never the path, which is where a Slack webhook keeps its credential. That detail is what makes the boot log safe to add at all.

I checked out the branch and ran it rather than reading the diff:

- `bun run typecheck` in `server/` — clean.
- `bun test test/egress-destinations.test.ts test/pr-nudge.test.ts test/alerts-sink.test.ts test/no-raw-exception-to-the-caller.test.ts test/readme-remote-claims.test.ts` — **30 pass, 0 fail**.
- `bunx oxlint` on the touched files — one warning at `alerts.ts:169`, pre-existing and not yours.

CI is green too, now that I've approved the run: `build`, `mobile`, both CodeQL analyses, GitGuardian.

Three things came out of the review, and **all three are mine rather than yours — I've filed them as #562 and I'll do them myself. Nothing here is waiting on you.** Writing them down anyway, because two of them are about your code and you should see them here rather than find them in a later diff.

**1. The base-URL guard needs to be consulted one level up**

`anthropicClientOptions()` is called from `viaApi` and nowhere else. `generateWalkthrough` tries `viaClaudeCli` first whenever `claude` is on PATH, and that spawn inherits the environment whole (`env: { ...process.env, AGENTGLASS_INTERNAL: "1" }`, `server/src/walkthrough.ts:177`). The CLI reads `ANTHROPIC_BASE_URL` itself, so on the path that runs by default the diffs still go wherever that variable points — when the CLI path succeeds, the new guard is never consulted at all.

The boundary you built is the right place for the check; it just has to run before the transport is chosen rather than inside one of them. I'll move it there and pin the spawn env in `server/test/walkthrough-egress.test.ts`, which already guards what this feature may send.

**2. Trusting the hosts the webhook exists for — my call, not a defect in yours**

`webhookDestination()` passes an empty `trustedHosts`, so `hooks.slack.com` and Discord fall under `AGENTGLASS_ALLOW_REMOTE` too. That is a fair reading of #468, which named the shape and never said which hosts stay trusted. It is also a behaviour change for every install that already has a webhook, since the variable exists for Slack or Discord in the first place. The distinction I wanted and failed to write into the issue: the webhook carries notification text, the walkthrough carries repository code. So I'd trust those two hosts — `trustedHosts` already exists for exactly that — and keep the hard refusal for `ANTHROPIC_BASE_URL`. That one is on me, not on you.

**3. SECURITY.md still describes the world before the kill switch**

`docs/CONFIG.md` is updated, but SECURITY.md has three sentences this PR makes untrue: the *Hardening knobs* table (~714) doesn't list `AGENTGLASS_WALKTHROUGH_DISABLED`; "Two things are **not** individually switchable … the AI **Explain** walkthrough" (~735); and "it is the only one that does so unasked by a switch" (~359). *What leaves through the webhook* (~690) should also say a remote destination now needs the opt-in. I pin doc claims in tests here (`readme-remote-claims`, `retention-promises`, `walkthrough-egress`), so prose is treated as code in this repo — nothing you could have known from outside, and I'll write those paragraphs.

A few small things are in #562 too, purely for the record: the `AGENTGLASS_ALLOW_REMOTE` row in `docs/CONFIG.md` lost the sentence explaining *why* it is deliberately awkward (a repo-local `settings.json` can set `AGENTGLASS_SERVER`) and I'd like that back; `LOOPBACK_HOSTS` covers `127.0.0.1` but not the rest of `127.0.0.0/8`, the same gap `hooks/send_event.py` has; and `alerts.ts` resolves the destination once at import while `prnudge.ts` re-reads it per call, which is harmless but worth a comment saying which is intended.

The template reminder above is a bot, and your validation notes already say more than the template asks for. Thank you for picking up a deferred security issue rather than an easy one — this is the second time the egress paths in this app have been tightened from outside, and both times what improved was the shape of the thing, not just the code.

---

PR #564, opened 2026-09-06
fix(theme): confine machine sync to explicit picks

### LINKED ISSUE #455: Theme sync repaints the user's own tmux, and a headless page load can trigger it
### What happens

Picking a theme in any cockpit UI repaints the user's own terminals. `POST /theme/sync` writes `~/.config/agentglass/theme.tmux.conf` and sources it into the **default** tmux socket, which is where a normal `tmux` on that machine lives. So the palette does not stay inside the app: it lands on every pane of every personal session, plus nvim through `theme.lua`.

Seen today on my machine. Both servers came up white:

```
socket default    (my own session, 4 windows) -> window-style "bg=#ffffff"
socket agentglass                             -> window-style "bg=#ffffff"
```

`~/.config/agentglass/theme.tmux.conf` was stamped `# Theme: Light`, written at a moment when nobody was sitting at the picker. The OS was `prefer-dark` and the desktop theme had been set to a dark one 41 minutes earlier, so this was not the System-follows-the-OS path in `watchSystemScheme`. The most likely trigger is one of the headless harnesses that drive the app: they load the UI in headless Chrome, and anything that reaches `pickTheme` there broadcasts machine-wide.

The failure is silent and looks like a renderer bug. We have chased it twice before as a WebGL/xterm white-out (#280) and once as a test writing a Light palette into the real home (#315). Both of those are fixed; this is a third, structural route to the same white screen.

### What should happen

A theme picked in a browser tab is per-browser state. Carrying it out to the machine should require intent, and should never be something an automated page load can do.

Concretely, some mix of:

- Only sync when the pick came from a real user gesture, not from any programmatic `applyTheme`/`pickTheme` call.
- Do not write the default tmux socket implicitly. Prefer the app's own socket, and treat the user's personal server as opt-in.
- Refuse the push when the request comes from a headless/automated client, the same shape as the `NODE_ENV=test` fence already in `themesync.ts`.
- Keep a backup of what we overwrote so a bad push is one command to undo.

### Where

- `web/src/lib/themes.ts` (`pickTheme`, `applyTheme`, `syncTheme`, `watchSystemScheme`)
- `server/src/themesync.ts` (writes the conf, sources it live, pushes to nvim)
- `server/src/index.ts` (`POST /theme/sync`)

### Repair, for anyone hitting it now

```
tmux -S /tmp/tmux-1000/default set -gu window-style
tmux -S /tmp/tmux-1000/default set -gu window-active-style
tmux -S /tmp/tmux-1000/default set -gu status-style
tmux -S /tmp/tmux-1000/default source-file ~/.config/tmux/tmux.conf
```

Related: #280 (WebGL white-out, renderer by platform), #315 (test suite wrote a Light theme into the real home), #339 (theme sync owning the user's status bar).

be-student:
Closes #455.

Theme restoration and OS changes now stay browser-local, automated browsers cannot cross the sync boundary, and live tmux reload explicitly targets AgentGlass's configured named socket. Explicit picker actions still update AgentGlass's tmux and nvim integrations.

Validation:
- 12 focused theme tests
- server and web type checks
- lint and production build
- performance budget


SirAllap:
Thank you for taking this one, @be-student. #455 had three separate routes tangled in it, and you closed the two that actually produced the white-out I filed it for: a page load that broadcasts, and a bare `tmux` that lands on whatever socket `$TMUX` names.

I checked out the branch and ran it rather than reading the diff:

- `bunx tsc --noEmit` in `server/` — clean.
- `bun test server/test/themesync-nondestructive.test.ts web/test/theme-boot-sync.test.ts` — **12 pass, 0 fail**.
- `bun test server/test/themesync.test.ts web/test/theme-mode.test.ts web/test/theme-repair.test.ts` — **58 pass, 0 fail**.
- `bunx oxlint` on the touched files — clean.

CI is green too: `build`, `mobile`, both CodeQL analyses, GitGuardian.

The part I appreciated most is that naming the socket does more than pick a target. `socketPath()` (`server/src/tmuxctl.ts:1805`) resolves `["-L", name]` to the engine socket, so `tmuxSocketAllowed` and `tmuxSocketConfined` now judge the socket that will actually be written instead of one they cannot see — on the bare-`tmux` path those guards were blind, and the comment you replaced admitted as much. Routing through `resolveTmuxBin()` matters for a reason that is easy to miss from outside, too: `tmuxbin.ts:130` probes a server already running on that socket so the client's protocol version matches it. A bare `tmux` there would have brought back "server exited unexpectedly" on exactly the machines that keep a warm engine.

Two things came out of the review. **Neither is a defect in your code, and the second one is mine — I never wrote down which shape I wanted.**

**1. The terminal panel still repaints the user's own server, with no pick involved**

`followSession` calls `applyThemeTo(t.socket)` at `server/src/terminal.ts:1047`, every time the panel follows a tmux session. That socket comes from `socketOf()` off a real client's argv — for both of my own clients it is `[]`, which is precisely the bare spelling that resolves to my personal server — and it sources the same generated `theme.tmux.conf`. So "I attached my own session in the panel and it went white" is still reachable after this diff, and now it is reachable with a *stale* palette, because `syncTheme` has stopped keeping that file current for that socket.

Small and related: the docstring above `applyThemeTo` (`server/src/themesync.ts`, just below `tmuxConfPath`) still says *"`syncTheme` already does this on the default socket when the palette changes"*. That sentence is the justification for the function existing, and it is no longer true.

**2. `navigator.webdriver` does not catch this repo's own harnesses — my miss, not yours**

I probed it rather than assuming. Launching Chrome the way every script in `scripts/` does — `--headless=new --remote-debugging-port=… --user-data-dir=…`, then driven over CDP (`scripts/cdp.ts`, `scripts/smoke.ts:113`, `scripts/browser-drive.ts:93`, `scripts/eyes.ts:143`) — reports:

```
navigator.webdriver = false
```

That flag is set by WebDriver and by libraries that pass `--enable-automation`; a plain CDP attach is neither. Since #455 named those harnesses as the most likely trigger, the client-side check is a good first line but not the fence. The issue asked for the refusal server-side, next to the `NODE_ENV=test` fence in `themesync.ts`, and that is the one a harness cannot step around. Since I asked for it vaguely, I'm happy for it to be a follow-up rather than a change here.

So: I'd like to merge this, and I'd like to **keep #455 open** rather than have it close on merge — the panel route above, the server-side fence, and the fourth bullet in the issue (keep a backup of what we overwrote) are all still outstanding. Either drop the `Closes #455` line and I'll re-scope the issue to what's left, or say the word and I'll do that after merging.

The template reminder above is a bot; your description already says more than the template asks. Thank you for picking up a bug that needed three reads of the socket rules before it could be fixed at all.

---

SirAllap:
Merged. Thank you, @be-student — this is the second structural fix you've landed here in a week, and both times what improved was where the boundary sits rather than the code either side of it.

I left the `Closes #455` line alone, so the issue closed with the merge — same as #468 did on #561. What's left of it is in #567 along with the perf-harness gap: the panel's own repaint through `applyThemeTo`, the server-side refusal a CDP-driven Chrome can't step around, and the backup bullet. **All three are mine. Nothing is waiting on you.**

---

Issue #109, opened 2026-07-20
Gate: rule-based auto-hold (spend threshold / tool allowlist), not only human approval


SirAllap:
## The gap

The gate today is **reactive and human-driven**: a `PreToolUse` hook holds a call until a person clicks approve/deny in "What needs you" (or the timeout auto-allows). That only catches things while someone is watching.

But the failure mode that actually costs money is a **runaway loop that spends before anyone looks** — by the time the dashboard shows the cost spike, the money is already gone. The live view documents the problem faster; it does not prevent it. And a per-matcher gate that fires on every `Bash` is coarse: it either nags on every call or gates nothing.

## Proposal — let the gate hold/deny by rule, no human needed

Add automatic policy evaluation inside the gate so a call can be held or denied without waiting for a click:

- **Spend threshold** — hold (or deny) a tool call once a session's cumulative cost, or a projected step cost, crosses `$X`. The cost data already exists (`server/src/pricing.ts` + the per-session totals in `server/src/db.ts`), so the gate can consult it.
- **Tool allowlist / denylist** — auto-allow tools on a list, hold or deny anything outside it. Even a dumb static allowlist catches the "agent went rogue and started running things it never should" case.
- **Auto-hold vs auto-deny** — a soft rule surfaces the call in "What needs you" (human still decides); a hard rule denies outright with a reason.
- **Scoped per project / agent, not global** — the common failure is *one* workflow going rogue while the rest of the fleet is fine, so a blanket kill switch is too blunt. This shares the per-root policy shape proposed in #14.

## Why it matters
This turns the gate from "a faster way to document a spend spike" into "the thing that stops the runaway loop before it becomes a bill" — which is the whole point of having a control plane rather than just a viewer.

## Entry points
`server/src/gate.ts` (decision path), `hooks/gate_event.py` (the hook), the policy/config surface in `~/.config/agentglass/config.json` (align with #14), and the cost figures in `server/src/pricing.ts` / `server/src/db.ts` for the threshold rule.

Related: #14 (per-project gate policy), #43 (gate durability).

Issue #14, opened 2026-07-19
Per-project governance: gate policies and hook profiles scoped to a repo


SirAllap:
Every safety control agentglass has today is global: the approve/deny gate, fail-open vs fail-closed (`AGENTGLASS_GATE_FAILCLOSED`), the write-disables. But risk isn't global — the same person wants "everything needs my approval" on the production repo, "anything goes" in the sandbox, and "no pushes, ever" on a client's codebase. Today the only way to express that is restarting the server with different env vars. Per-project governance + per-project hooks is how the serious setups run.

## Proposal

A **policy per project root**, resolved by longest path match against a session's cwd:

```jsonc
// ~/.config/agentglass/config.json
{
  "policies": {
    "~/code/prod-api":   { "gate": ["Bash", "Write"], "failClosed": true, "gitWrite": false },
    "~/code/sandbox":    { "gate": [] },
    "~/clients/acme":    { "gitPush": false, "chatBypass": false }
  }
}
```

- The **gate** (`server/src/gate.ts` + `hooks/gate_event.py`) consults the policy for the calling session's project instead of one global switch — per-project tool matchers, per-project fail-open/closed
- The write-gates (git push, docker, chat bypass) become overridable per root, with the env vars keeping the global default
- **Hook profiles**: `hooks/install_hooks.py --project` already writes per-repo hooks; a policy could declare which profile a project gets (full gating vs telemetry-only), so `bun run setup` wires each project according to its declared risk
- UI: show the active policy on the session card / gate prompt ("gated by prod-api policy"), so an approval request explains *why* it's asking

## Why it matters

The gate is agentglass's most differentiated feature, and one-size-fits-all is what keeps it turned off. Policies that match how people actually rank their repos ("prod is sacred, sandbox is free") make the control plane something you'd leave on.

PR #565, opened 2026-09-06
feat(insights): price expired cache rebuilds

### LINKED ISSUE #293: [feat]: price the cache you paid to rebuild
### What problem would this solve?

Prompt caches expire. When a turn resumes after the window has passed, the whole prefix is re-encoded at the write rate instead of read at the cheap one, and the difference is real money that never appears as a line item. It looks like ordinary spend. As someone watching a fleet I can see that a session cost a lot, but not that part of it was paid twice for the same context.

### Proposed solution

A cost insight that detects and prices avoidable re-encodes. We already ingest `cache_creation_tokens`, `cache_read_tokens`, a timestamp per turn, and whether the turn belongs to a subagent. Partition a session's turns by a lineage proxy (model plus main/subagent), then flag a large `cache_creation` that lands after a gap longer than the cache window on a lineage that previously had an established prefix. The penalty is exactly `cache_write - cache_read` per token, which `pricing.ts` already knows.

Surface it as a first-class insight next to the existing spend and burn kinds: "rebuilt cache you already had, N times, $X." A predictive variant could warn while a cache is still alive and close to expiring.

Every figure should be computed from the user's own events, so the number always refers to their machine and their sessions.

### Alternatives considered

Nothing today separates a re-encode from ordinary input in the cost charts, so the money is invisible by construction.

be-student:
Closes #293.

Adds a 24-hour cache-rebuild insight for large creations after the five-minute cache window. Detection stays within the same session, model, and main/subagent lineage, and prices only the cache-write premium over a read.

Validation:
- 20 focused insight and scope tests
- server and web type checks
- lint, production build, and performance budget
- web suite: 4,274 passed; 23 unrelated existing keyboard/layout failures
- server suite: changed tests passed; unrelated host/tmux fixtures failed
- smoke mounted the app; the existing workspace-shortcut invariant failed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ Where this is going — themes, not dates. The living version is the issue track
 - Tasks per project, and a decision log mined from transcripts — [#12](https://github.com/SirAllap/agentglass/issues/12), [#13](https://github.com/SirAllap/agentglass/issues/13)
 - Voice input in chat — [#92](https://github.com/SirAllap/agentglass/issues/92)
 
+**Built since v0.15.0**
+- Cache rebuild insights identify a large prompt cache recreated after expiry, keep model and main/subagent lineages separate, and price only the cache-write premium over a cache read — [#293](https://github.com/SirAllap/agentglass/issues/293)
+
 
 ## Released
 
--- a/server/src/insights.ts
+++ b/server/src/insights.ts
@@ -3,7 +3,7 @@
 // Computed from the full event history (better than the live buffer for loops).
 import type { Insight } from "../../shared/types.ts";
 import { db, scopeClause } from "./db.ts";
-import { equivalentTokens } from "./pricing.ts";
+import { cacheRebuildPenaltyUsd, equivalentTokens } from "./pricing.ts";
 import { budgetStatus, budgetScopeLabel, periodLabel } from "./budget.ts";
 
 const key = (app: string, sid: string) => `${app}:${sid.slice(0, 8)}`;
@@ -14,6 +14,14 @@ const key = (app: string, sid: string) => `${app}:${sid.slice(0, 8)}`;
 // `b:c` are different rows that a plain join hands the same string.
 const rowId = (app: string, sid: string) => `${encodeURIComponent(app)}:${encodeURIComponent(sid)}`;
 const trim = (s: string, n = 64) => (s.length > n ? s.slice(0, n) + "…" : s);
+const USD = new Intl.NumberFormat("en-US", { minimumFractionDigits: 2, maximumFractionDigits: 2 });
+
+// The bundled cache-write rates describe five-minute prompt caches. A smaller
+// creation is usually normal turn churn; keep this signal for prefixes large
+// enough to be useful rather than announcing pennies on every resumed chat.
+export const CACHE_REBUILD_WINDOW_MS = 5 * 60_000;
+export const CACHE_REBUILD_LOOKBACK_MS = 24 * 60 * 60_000;
+export const CACHE_REBUILD_MIN_TOKENS = 10_000;
 
 export function getInsights(): Insight[] {
   const now = Date.now();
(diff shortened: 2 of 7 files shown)


SirAllap:
Thank you for this, @be-student. The inference in this card is the right one, and the tests are the tests I would have asked for: keeping model and main/subagent lineages apart is what stops it from pricing a model switch as waste; accepting a prior cache *read* as evidence the prefix existed is the case I would have missed; `p.model_name IS e.model_name` rather than `=` is the NULL-safe comparison this table needs; and adding the cache lineage to `insights-scope.test.ts` shows you went looking for the cross-project leak that file exists for, instead of waiting for me to find it.

I have to hold it anyway, and it is a measurement rather than an opinion.

**`/insights` becomes unanswerable on a real database**

The correlated subquery runs once per candidate row, and `EXPLAIN QUERY PLAN` says what that costs:

```
SEARCH e USING INDEX idx_events_ts (timestamp>?)
CORRELATED SCALAR SUBQUERY 1
SEARCH p USING INDEX idx_events_model_cov (model_name=?)
```

Per candidate it walks every event sharing that model name — on a desk that runs one model most of the day, close to the whole table. I filled a scratch DB with cache-bearing turns spread over 24 hours and timed `getInsights()` on `main` and on this branch:

| events (sessions × turns) | `main` | this branch |
| --- | --- | --- |
| 1,000 (5 × 200) | 0.6 ms | 60 ms |
| 5,000 (10 × 500) | 1.7 ms | 1,330 ms |
| 30,000 (20 × 1,500) | 11 ms | **46,000 ms** |

Two things make that worse than the table alone. `bun:sqlite` is synchronous and `/insights` is served inline (`server/src/index.ts:3136`), so those 46 seconds are 46 seconds in which the whole server answers nothing — loopwatch would record it as the process having gone away. And `Alerts.tsx` polls it every fifteen seconds (`usePoll(active, loadInsights, 15_000)`), so the next request arrives long before the previous one is done. The 30,000-event run produced **no cache cards at all**: the cost is paid whether or not anything is flagged.

None of this is something your validation could have shown you, and that part is mine: `scripts/perfbudget.ts` measures event-loop delay against a fresh database, so the one guard I have for exactly this failure is blind to it. Retention is 8 days, so a real install carries far more than 30k rows. I'll widen the budget script separately.

A single pass with a window function does the same job. I ran it against the same 30k dataset — **111 ms**, and it narrows further if the CTE is restricted to sessions that have a candidate:

```sql
WITH turns AS (
  SELECT source_app, session_id, model_name, cache_creation_tokens, timestamp,
         LAG(timestamp) OVER (
           PARTITION BY source_app, session_id, model_name, COALESCE(NULLIF(agent_id,''),'main')
           ORDER BY timestamp, id) AS previous_at
    FROM events
   WHERE timestamp > ? AND (cache_creation_tokens > 0 OR cache_read_tokens > 0)
)
SELECT * FROM turns
 WHERE cache_creation_tokens >= ?
   AND previous_at IS NOT NULL
   AND timestamp - previous_at > ?
```

`PARTITION BY model_name` groups NULLs together, so it keeps the property your `IS` was there for. One behaviour difference worth deciding deliberately rather than inheriting: the prior turn now has to fall inside the lookback window, so the first candidate after a 24-hour silence is not flagged. I think that is the better answer — a cache that expired yesterday is not news — but it should be a choice with a test on it.

**Two smaller things**

- **No cap.** Every other block in `insights.ts` limits itself (`LIMIT 6` at :47, `LIMIT 4` at :67 and :154). This one emits a card per session with any qualifying rebuild in 24 hours, so a busy fleet fills the panel with them. A `LIMIT` ordered by penalty would match its neighbours.
- **CHANGELOG.** "Built since v0.15.0" is a new shape for that file. The convention here is a `**vNEXT (unreleased — the number is the release commit's to set)**` bullet under `## Released`, which the release commit then renames — 35f9cb5 is the last one doing it.

Everything else I'd take as it stands: the thresholds, the aggregation, the `cacheRebuildPenaltyUsd` premium over a read, the icon. Take the rewrite if you want it — it's yours, and I'd rather this land under your name than mine. If you'd prefer I push the query onto the branch so the shape is settled and you keep the tests, say so and I will.

One last thing to watch, since it lands the same week: #566 maps an absent `cache_creation_input_token_cost` to zero, and for a model that ends up priced that way `cacheRebuildPenaltyUsd` returns 0 and this card quietly stops appearing. I've raised it over there — flagging it here so whichever merges second doesn't hide the other.

---

SirAllap:
Re-reading my own comment, I left you with a question instead of an answer, which is a poor way to leave a PR. So, plainly:

**The rewrite is yours, and I'd rather it stayed yours** — the inference and the tests here are the hard part and they're already right. Take the time you need; there's no deadline on my side.

If you'd rather not, that is equally fine and costs you nothing: say so, or just go quiet for a day, and I'll push the query onto your branch myself (maintainer edits, or a patch if that is off) and leave your tests and your commit standing.

Two things settle it either way, whichever of us writes it: the prior turn now has to fall inside the lookback window — a cache that expired yesterday is not news — and the block wants a `LIMIT` ordered by penalty like its neighbours.

Nothing on #566 is waiting on me either: the mapping of an absent cache rate is the one change I'd want before merge, and the SECURITY.md paragraphs there are mine to write once it lands.

---

SirAllap:
Merged, as #570 — your commit, with the query rewrite on top of it.

I said I'd take it if the branch went quiet for a day, and it did, so I have. That is a scheduling decision and nothing else: **the substance of #570 is yours.** The inference is yours, the lineage boundaries are yours, and your twelve tests are in `main` untouched — which is exactly what made the rewrite safe to do, because "the same twelve assertions still hold" is the only evidence that changing the cost didn't change the meaning. Your commit `8f6d66d` is in the history with your name on it.

What I changed: `LAG` over the same partition instead of the correlated lookup — **46,000 ms → 79 ms** on 30k events — plus the cap at six cards, dearest first, and the CHANGELOG moved to the `vNEXT` bullet. The behaviour difference we discussed is now pinned by a test: the prior turn has to fall inside the lookback window, so the first candidate after a full day of silence is not flagged.

`make perf` catches this class of thing now (#568): it seeds 20k events and polls `/insights`, and it failed your branch at 1,410 ms with the watchdog naming the route. That gap was mine, not yours — your validation could not have shown it.

#566 has gone the same way, as #571, with the reasons in a comment there. Thank you for both of these; if you want another, #109 and #14 are the ones I'd most like help with.

---

PR #566, opened 2026-09-06
feat(pricing): refresh rates from LiteLLM

### LINKED ISSUE #9: Pricing table: keep model prices fresh without hand-editing pricing.ts
Cost tracking is only as good as `server/src/pricing.ts`, and model prices change (and new models ship) monthly. Today the README honestly says: *"verify current rates with each provider and tune"* — which means everyone's dashboard silently drifts wrong.

## Proposal

- An **opt-in refresh** from a community-maintained source — LiteLLM's [`model_prices_and_context_window.json`](https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json) is the de-facto standard (per-token input/output/cache pricing for every major provider)
- Fetch on boot (and daily), map into the existing pricing shape, **fall back to the bundled table offline** — the server must keep working with zero network
- Show provenance in the Stats modal ("prices: bundled · updated 2026-07-01" vs "live · litellm")
- `AGENTGLASS_PRICING` (user override file) keeps the last word, as today

## Why it matters

Cost-per-session is one of the headline features; making its numbers trustworthy by default removes the biggest asterisk on it.

be-student:
## What does this PR do?

Fixes #9. `AGENTGLASS_PRICING_REFRESH=1` now loads validated exact-model rates from LiteLLM at boot and daily, while bundled prices remain the offline fallback and user overrides remain authoritative. Statistics shows the active price source and update date.

## How was it tested?

Focused pricing tests (77 pass), all three typechecks, lint, production build, a live LiteLLM fetch, an isolated `/stats` smoke, and the performance budget (p99 5ms / 120ms) pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Where this is going — themes, not dates. The living version is the issue track
 - Per-project gate policies and hook profiles — [#14](https://github.com/SirAllap/agentglass/issues/14)
 - Keep model prices fresh without hand-editing the table — [#9](https://github.com/SirAllap/agentglass/issues/9)
 
+The optional LiteLLM price refresh now updates exact model rates at boot and daily, keeps bundled offline fallback, preserves user overrides, and reports its provenance in Statistics.
+
 **Later / exploring**
 - An API panel to exercise the endpoints the fleet is building — [#170](https://github.com/SirAllap/agentglass/issues/170)
 - Tasks per project, and a decision log mined from transcripts — [#12](https://github.com/SirAllap/agentglass/issues/12), [#13](https://github.com/SirAllap/agentglass/issues/13)
--- a/docs/CONFIG.md
+++ b/docs/CONFIG.md
@@ -114,6 +114,7 @@ are:
 | `AGENTGLASS_SCAN_DISABLED` | — | `1` → turn off the machine-wide transcript scanner (rely on hooks / OTel only). |
 | `AGENTGLASS_RETENTION_DAYS` | `8` | Days of **raw events** to keep (pruned hourly, with resolved gates and answered reminders). Covers the full 7d stats window. Expiring days are folded into a daily rollup first, so spend history outlives the rows. `0` = keep events for ever, and also keep the Clone's shifts, acts and runs for ever. It does **not** move the Clone's and Lantern's own windows (30 and 90 days, fixed) — see SECURITY.md, *What agentglass stores*. |
 | `AGENTGLASS_PRICING` | — | Path to a JSON pricing override (see `server/src/pricing.ts`). |
+| `AGENTGLASS_PRICING_REFRESH` | — | `1` (or `true`/`yes`/`on`) → refresh exact model prices from LiteLLM at boot and daily. The bundled table remains the offline fallback; `AGENTGLASS_PRICING` disables the refresh and stays authoritative. |
 | `AGENTGLASS_WEBHOOK` | — | POST `{text}` here (a Slack/Discord-shaped incoming webhook) for alerts, Lantern watch notices and PR nudges sent with `send: true`. Loopback and the two hosts it exists for (`hooks.slack.com`, `discord.com`) work directly; any other destination requires `AGENTGLASS_ALLOW_REMOTE=1`. The host it resolved to is logged once at boot — the host, never the path, which is where a webhook URL keeps its credential. What travels is the notification's title and body — agent names, a checkout's base name, a PR's number, title, URL and reviewer logins — never a transcript, a diff or a token. Blanked out of every Clone run. |
 | `AGENTGLASS_NOTIFY` | — | `1` → fire desktop alerts. A connected client (browser or desktop app) raises a **native OS notification** on any platform; `notify-send` is the fallback for a headless server with nothing attached to show it. Does **not** gate the phone, which hears every alert on the socket it already holds and decides for itself — see [Alerts on the phone](WORKSPACE.md#alerts-on-the-phone-and-what-they-honestly-cover). |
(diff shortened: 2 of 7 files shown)


SirAllap:
Thank you for this, @be-student. #9 has been on the roadmap since the price table was first written, and the shape you chose is the one I wanted without having said so: the bundled table as the offline floor, live rows only for an exact model id, the user's override still authoritative over both, and the whole thing off unless somebody asks for it. `installLive` being all-or-nothing behind a minimum size is the detail that makes it safe — a half-downloaded catalogue can't become the price of everything.

I checked out the branch and ran it:

- `bunx tsc --noEmit` in `server/` and `web/` — clean.
- `bun test server/test/pricing-refresh.test.ts server/test/pricing.test.ts` — **22 pass, 0 fail**.
- `bunx oxlint` on the touched files — clean.
- Fetched the real catalogue: **HTTP 200, 2.31 MB, 3,818 entries**, 3,186 of them with usable input and output rates — comfortably inside both the 5 MB ceiling and the 100-model floor.
- Checked the exact-match assumption against names we actually record: `claude-opus-4-8`, `claude-sonnet-4-5-20250929`, `claude-opus-4-1-20250805`, `gpt-5.1-codex`, `gemini-2.5-pro`, `claude-haiku-4-5` — all hit. The stripped `[1m]` suffix is handled, and your test pins it.

CI is green: `build`, `mobile`, both CodeQL analyses, GitGuardian.

Four things came out of the review. The first is the one I'd want changed before merge; the rest are mine or close to it.

**1. An absent cache rate is read as "free", and it silently beats a number we verified**

`perMillion(value, true)` returns 0 when the field is missing, so a live row with no `cache_creation_input_token_cost` prices cache writes at zero — and because a live exact match wins over the bundled table, that zero replaces a rate we had checked by hand.

In today's catalogue **2,801 of the 3,186 usable entries carry no cache-creation cost**. Concretely: `minimax.minimax-m2` has a cache *read* price and no cache *write* price, so our bundled `cache_write: 0.375` becomes `0`. Missing is not free — for most of those providers a cache write is billed at the input rate.

I'd rather an absent field fell back to the bundled row's value for that field, and to the input rate where there's no bundled row, keeping zero only when the catalogue actually says zero. Same for `cache_read_input_token_cost`. It also matters beyond the totals: #565 adds a card priced on `cache_write - cache_read`, so a model whose write silently becomes free loses the insight altogether.

**2. It goes around the boundary you built last week — my call rather than a defect**

`refreshLiteLlmPricing` calls `fetch` directly. Strictly it doesn't need `outboundDestination()`: the URL is a constant and the feature is off by default. But that function is now the single place that answers "may this process talk to that host", and a call that skips it is a call that meets no gate on the day someone edits `LITELLM_PRICING_URL` or points it at a mirror. `raw.githubusercontent.com` as a trusted host, in the same shape as `WEBHOOK_SERVICE_HOSTS`, costs a line and keeps the answer in one place. `redirect: "error"` and the streamed size ceiling are both right, and I'd keep them either way.

**3. SECURITY.md still describes the world before this refresh — mine, I'll write it**

`docs/CONFIG.md` is updated and the row you added is good. SECURITY.md is the file that enumerates outbound requests, and it now under-counts: *"One more outbound request: assignee pictures"* (~345) and *"Two more paths exist and both are off until you turn them on"* (~365). This is a third, and it fires at boot and every 24 hours. Prose is pinned by tests here (`readme-remote-claims`, `retention-promises`, `walkthrough-egress`), so I treat it as code — the same gap I flagged on #561, and nothing you could have known from outside.

**4. Two honesty nits on the provenance chip, and the CHANGELOG**

`updated_at` is the date we fetched, not the date the catalogue changed, so the chip reads "updated <today>" after pulling a table that's a month old; the user-table branch dates itself from process start for the same reason. And `cost_usd` is computed at ingest, so a refresh reprices nothing already stored — the chip sits above a Statistics window whose older rows were priced by whatever was loaded then. A word in the label or in CONFIG.md would keep it honest.

On the CHANGELOG: the paragraph landed inside "Where this is going" while #9 stays listed under **Next**, so the file now says the thing is both shipped and upcoming. The convention is a `**vNEXT (unreleased — the number is the release commit's to set)**` bullet under `## Released` (35f9cb5 is the last release commit renaming one), and the roadmap bullet should go.

For the record on ordering: I trial-merged this and #565 onto `main` in both orders and they go in clean, even though both touch `pricing.ts` and `CHANGELOG.md`. Nothing here is waiting on that one.

---

SirAllap:
Merged, as #571 — your commit, with the three changes from my review on top. Same as #565: I said I'd take these if the branches stayed quiet for a day, they did, and the scheduling is the only thing that decided it. The shape of this feature is yours and I kept every part of it — live rows only for an exact model id, the bundled table as the floor, `AGENTGLASS_PRICING` authoritative over both, `installLive` all-or-nothing behind a minimum size, the streamed size cap, `redirect: "error"`. Your twenty-two tests are in `main` unchanged.

What I changed, and the middle one is the reason I couldn't just merge it:

**An absent cache rate now reads as the input rate, not as free.** I checked this against the live catalogue rather than reasoning about it: 843 rows publish a discounted `cache_read` and no `cache_creation` — OpenAI's and Gemini's shape, where a cache write costs the ordinary input rate and only the read is cheaper — and 1,971 publish neither. Zero would have replaced hand-verified rates with nothing and understated every cached turn on those models, and since `cost_usd` is computed once when an event is stored, those numbers would have stayed wrong. The 37 rows with an explicit `0` keep it: meaning it is different from not saying it.

**The fetch goes through `outboundDestination()`** with GitHub's raw host trusted for this one read — the boundary you built in #561, which I'd rather nothing routed around, however safe the constant is today.

**The provenance date says what it knows.** "As of" rather than "updated", since for a live catalogue the date is when we fetched; the user's table now dates from its own mtime instead of from process start.

Plus the prose that was mine to write and that I flagged as mine: SECURITY.md enumerates every outbound request and now names this one, spelling out that it carries no credential and nothing about you; CONFIG.md says the refresh prices new events only.

That's three of yours landed this week — #561, #570 and #571 — and the pattern I keep noticing is that you pick the deferred ones. Thank you. If you want another, #109 and #14 are where I'd most value the help.

---

PR #594, opened 2026-09-13
feat(terminal): Ctrl, Alt and Shift as latches on the phone's key bar

SirAllap:
## What does this PR do?

The key bar could send any combination somebody had written into a table in advance, and nothing else. **That ceiling is visible from inside the app**: an agent's permission modes cycle on `Shift+Tab`, so a bar carrying `Tab` and no `Shift` is being asked, on screen, for a key it does not have.

### The encoder was already here

`keyBytes` has handled Ctrl, Alt and Shift over cursor keys, control codes and Alt-as-Escape since the tmux prefix needed it — and only the prefix ever called it. What was missing was a way to say *which modifiers are down* on a surface where nothing can be held down.

### So they latch

| taps | state | drawn as |
|---|---|---|
| one | armed for the next key | primary outline, primary label |
| two | locked until tapped off | outline **and** the pressed background |
| three | off | an ordinary key |

A key press spends the first and not the second. A combination pressed several times in a row is exactly the case that made the bar feel short, and re-arming between each press is the tax that stops people using a feature.

No new fill and no new colour: `primary` is a foreground everywhere else on that screen and `bg4` is what a press already looks like. A latch is a state of a key here, not a fourth kind of control.

### Refusing beats ignoring

A key with no encoding for the combination — a control code, which is *already* a Ctrl press, or a macro, which is text — sends nothing and is drawn unavailable while the latch is up. Sending it plain is the quiet failure: a Tab arriving on the line of somebody who pressed Ctrl first, with nothing to say the Ctrl went nowhere. This is the rule the encoder already applied to a character with no control code.

### Placement, and a bug this found

They sit one place behind the six the fold was measured for. Esc and Ctrl+C earn those places whether or not anybody taps a modifier, and paying for a new feature with the key that stops a runaway command is the wrong trade — anybody who lives on the latches moves them forward on the settings screen, where the same trip lets them delete the nine preset Ctrl+letter keys these replace.

Adding them exposed a real hole in `canHide`: it counted *keys*, and Ctrl/Alt/Shift are keys that send nothing, so a bar could be hidden down to nothing but latches — no Escape, no Ctrl+C, every press changing what the next press does and no next press to make. It now counts keys that **send**.

### The ceiling, said out loud

A latch reaches the keys on the bar, **not the letters on the soft keyboard** — so `Ctrl+K` is the preset key, not Ctrl and then K.

That is a deliberate stop, not an oversight. Typing goes through a field whose contents are diffed against what has already been sent; taking a character back out of that diff means rewriting the field mid-keystroke, which is the measured render race that once made this app send `hhehelhello` for `hello` and the reason the field stopped being cleared while anybody is typing. The nine preset Ctrl+letter keys already cover most of what that would have bought. What presets *cannot* cover is the modified navigation keys — `Ctrl+←`, `Ctrl+Home`, `Alt+⏎`, `Shift+PgUp` — and that is what this does reach.

## How was it tested?

`mobile/test/modifiers.test.ts`, 18 tests, plus the catalogue and layout tests extended:

- **The latch**: the three-tap cycle, one press spending an armed latch and sparing a locked one, stacking and spending independently, and that `NOTHING_HELD` is never written through (it is frozen and shared — a latch that armed itself would be a stray control code).
- **What a key sends**: `Shift+Tab` producing the same bytes as the preset `⇧Tab`, modified cursor keys against the encoder rather than against a hand-written expectation, Alt as a prefixed Escape, and control codes and macros refusing.
- **Table against encoder**: every key carrying a `key` name is checked to agree with `keyBytes` for four combinations — the way this goes wrong is one mis-named key that sends the wrong thing *only* while a modifier is up, which is the case nobody presses while checking.
- **The screen**: a source assertion that the bar still asks `sendFor`, because every test above can pass while the screen ignores the latch entirely — three buttons that light up and change nothing, which reads as broken rather than reverted. Comments are stripped first, or the scan would pass on the strength of the prose explaining it.

Every guard was broken on purpose and watched go red. **Two stayed green** and both are now covered: an unreachable second-line check in `sendFor` (pinned with a deliberately malformed key) and the `canHide` fix (pinned with a modifiers-only bar).

- `make lint`, `make typecheck`, and mobile's own `tsc` for **app and tests** — the test config caught two errors the app config did not.
- Mobile tranche: 826 pass / 41 skip / **0 fail**, against a measured baseline of 805 / 41 / 0 on the same tree without this change.


SirAllap:
`Analyze (javascript-typescript)` is red, and it is not this change. Recording why, rather than re-running until it is green.

**The analysis succeeds. The upload is what dies.** Both attempts end the same way:

```
Post-processing sarif files: ["/home/runner/work/agentglass/results/javascript.sarif"]
##[group]Uploading code scanning results
Uploading results
                              ← 18s of nothing
Post job cleanup.
```

Every query is interpreted and the SARIF is exported before that. There is no error message at all — the step hangs and the job dies.

**The control experiment is `Analyze (python)`.** It failed identically on the first attempt and passed on the re-run. This diff does not touch a line of Python, so whatever took it down was not in the diff — and the JS/TS SARIF is the much larger of the two, which is the one an unhealthy upload endpoint would keep refusing.

Two more data points from the same window: creating this pull request returned `500` three times (one of which had in fact created it), and the first `rerun-failed-jobs` call returned `502`.

Not portable, so nothing is being ported: there is no fix to carry for a code-scanning upload that will not accept a file. Every other check on this head is green — `build`, `mobile`, both `Sidecar` jobs, `nudge`, GitGuardian — and CodeQL last completed successfully on `main` at `707c311`, before today.

Leaving it red and watched rather than merging past it.

---

SirAllap:
Closing: this already landed. `main` carries it as `0c23841`, with `8ae0c3d` beneath it, and `mobile/src/terminal/modifiers.ts` is on `main` today.

The merge went through during the window on the 13th when GitHub's write path was failing intermittently, and the API answered that call with a `500` — so the commit was written and the pull request record was not. Three separate operations on that day returned a `5xx` for work that had in fact been done: creating this pull request, the CodeQL SARIF upload, and this merge.

Closing rather than merging, because merging a branch `main` already contains would be a second merge commit for nothing.

---

PR #595, opened 2026-09-13
fix(chat): a project switch left the composer opening chats in the last one

takara159:
## Summary

- `defaultCwd` (the fallback cwd for a new chat) is cached in one localStorage
  key shared across every project. A project switch reloads the page but not
  that key, so a stale cwd from the project just left survived into the new
  scope.
- Every chat opened against that stale cwd was then immediately filtered out
  of the sidebar by the workspace scope, since its cwd could never match. The
  composer and "+ New" looked dead — typing or clicking just silently opened
  another invisible, misrooted chat each time.
- Fix: once the freshly-scoped repo list comes back, discard the cached cwd
  unless it still names a repo in that list.

## Test plan

- [x] `bun run typecheck` (web/server/electron)
- [x] `bun test` in `web/` — the suites touching this path (chat-agent,
      checkout-picker, view-header-titles, running-panes, view-chrome,
      idle-polls-are-gated) pass; unrelated failures elsewhere trace to
      `navigator.platform` reporting "MacIntel" on a local macOS dev machine,
      reproduced identically on `main`
- [x] `bun scripts/smoke.ts` — the one pre-existing console error it flags
      reproduces identically on `main`, unrelated to this change
- [x] Built the desktop app (DMG) with the fix and confirmed switching
      between two projects no longer leaves the chat panel stuck


SirAllap:
The line this replaces is on `main` as written (`ChatPanel.tsx:823`), and the failure described is the nasty kind: nothing errors, the composer just quietly roots each new chat somewhere the scope then hides it.

I ran the predicate through its cases. It holds, including the one that usually gets this wrong:

| cwd | in scope? | result |
|---|---|---|
| `/home/u/proj-a` | yes | kept |
| `/home/u/proj-a/src` | yes, a subdirectory | kept |
| `/home/u/proj-a-otro` | **no** | replaced |
| `/home/u/viejo` | no | replaced |

That third row is the reason `replace(/\/$/, "") + "/"` is there rather than a bare `startsWith`, and it is right.

**One case changes behaviour beyond the bug.** When `gitRepos()` answers with an empty list, the old line kept whatever cwd was held; this one clears it to `""`, because `some()` over nothing is false. If an empty answer only ever means "no repos in this scope" then clearing is arguably more correct than keeping — but it is a second change riding along with the first, and it is worth being deliberate about which one it is.

**No test.** The whole decision here is a pure predicate — cwd plus a repo list in, a cwd out — so it is the shape this repository likes to pull out of a screen and assert on directly. Four lines of table would pin the prefix case above, which is the part a later refactor is most likely to quietly undo.

Neither is blocking; both are worth a follow-up.

---

SirAllap:
Merged as `2871efa` — and I led with the nitpicks and forgot the important bit: **thank you.** First contribution here, and a sharp one.

What I appreciated, specifically:

- You chased down a bug with no error message attached to it. The composer and "+ New" just *looked* dead while quietly creating chats nobody could see — that is the worst kind to track down, because there is nothing to search for.
- Ten lines. You found the one line that was wrong rather than rebuilding something around it.
- You got the prefix case right. `/home/u/proj-a-otro` is not inside `/home/u/proj-a`, and the trailing-slash handling is the reason it isn't — that is the detail a quick `startsWith` gets wrong and nobody notices for months.
- Your test plan separated the failures that were already there from the ones your change could have caused, and said which machine they came from. That saved me the work of deciding whether to believe it.

The test and the empty-list case are follow-ups, not things you owe. Say the word if you want them; otherwise they are written down.

Hope you send more.

---

PR #596, opened 2026-09-14
fix(security): a route's answer is not the exception's text either

Dev-next-gen:
## What does this PR do?

Three routes still answer with the text of a caught exception. The guard added in #560 looks for `error:` immediately followed by `String(e)`, and these three reach the same string one operator away from that shape, so it never saw them:

- `/view/image` — `error: ` + the exception, in one template literal
- `/walkthrough` — `String(e?.message || e)`
- `prAsset` — the whole `Response` body is `` `upstream: ${String(e)}` ``

I measured the first one because it is the one that throws on an ordinary bad day: `writeFileSync` into a directory the process cannot write throws `Error: EACCES: permission denied, open '<the absolute path>'`, and the route puts all of it in the body behind `could not write it: `. That is the path of a directory this server chose, going to a caller which on this app is a phone on the network — the flow `refused.ts` was written for.

All three go through `failed(where, e, said)` now, so the real error reaches this process's stderr and the caller gets a sentence.

The test grows a second rule to keep it that way, looked for where the answer is actually built (`json(` or `new Response(`) rather than after `error:`. I wrote its ceiling next to it: a bare `e.message` stays allowed, because `SourceRefused` and `IngestRefused` are returned exactly that way and their sentences are written in this repository — telling those from a `TypeError` needs to know the class, and a line of source does not.

On the one thing a reader might reasonably call a behaviour change: the body of two HTTP error answers changes text. The field, the shape and the status code do not, nothing in `web/` or `mobile/` matches on those strings, and the information that left is the information this repo already decided should not leave.

## How was it tested?

`cd server && bun test test/no-raw-exception-to-the-caller.test.ts` with only the test change applied — red, naming `index.ts:6716`, `index.ts:7613`, `prs.ts:3190`. Green with the three fixes.

Then the whole of it: `bun run lint` (0 errors), `cd server && bun run typecheck`, `cd server && bun test` — 5310 pass, 19 skip, 2 fail, against 5309/19/2 on `main` in the same checkout. The two failures are the same two before and after, and both are this machine rather than the tree: `gh-lookup-retries` wants a `gh` on PATH and there is none here, and one unnamed `beforeEach` hook times out.

I did not run `web`, the production build or the smoke: nothing here is in `web/`, `shared/` or the server↔web contract.

AI tools used


SirAllap:
Checked this against the tree rather than reading it. All three leaks are live on `main` today — `index.ts:6716`, `index.ts:7613`, `prs.ts:3190` — so this is fixing something real, and `/view/image` is the one that matters: `EACCES` puts an absolute server path in a body that on this app goes to a phone on the network.

**The new guard is a real guard.** I put each of the three leaks back, one at a time, and it went red on each, naming the exact line:

| put back | test says |
|---|---|
| `error: \`could not write it: ${String(e)}\`` | `index.ts:6717` |
| `error: String(e?.message \|\| e)` | `index.ts:7614` |
| `` new Response(`upstream: ${String(e)}`) `` | `prs.ts:3190` |

Worth saying, because a source-scanning rule that catches nothing passes just as green as one that works.

**One gap, and it is only in the rule — not in the fix.** It searches for the `json(` / `new Response(` and the `String(e)` on the *same line*, so the identical leak survives being wrapped:

```ts
return json({
  ok: false,
  error: `could not write it: ${String(e)}`,
}, 500);
```

Measured, not guessed: with that in place the suite is `3 pass, 0 fail`. Since the file already writes down its other ceiling — the deliberate `e.message` allowance — this one belongs next to it, so the next reader can tell a limit that was chosen from one that was missed.

Not blocking. The leak is open now and this closes it; the wrapped shape is a follow-up.

---

SirAllap:
Merged as `3f01f40` — and I owe you a better reply than the one I led with. I went straight to what I could poke at and skipped the part that matters: **thank you.** This is your first contribution here and it is a genuinely good one.

What stands out, specifically:

- You found three leaks the existing guard was written to catch and didn't. Noticing that a rule has a blind spot is harder than adding the rule was.
- You measured the case that actually happens — `EACCES` on a real bad day, putting an absolute server path in a body bound for a phone — instead of arguing it in the abstract.
- You wrote the ceiling of your own test next to it, and were straight about the two response bodies whose text changes. Volunteering the weak spot in your own patch is the habit that makes a reviewer trust everything else in it.

The wrapped-`json(` gap I mentioned is a follow-up, not a debt you owe. If you'd like to take it, it's yours; if not, it's written down now and someone will.

Hope to see you back.

---

Dev-next-gen:
@SirAllap Thank you. The line I am keeping is this one: "Volunteering the weak spot in your own patch is the habit that makes a reviewer trust everything else in it." That is the part I work hardest to get right, and it is rare to see a reviewer say it out loud rather than treat the admission as a weakness.

I will take the wrapped-`json(` follow-up. Your own measurement is the specification: with the wrapped shape in place the suite still reports `3 pass, 0 fail`, so the rule has to see past a single line. And you are right that the new limit belongs written down beside the deliberate `e.message` allowance the file already records, so that a ceiling which was chosen and one which was missed never look the same to the next reader.

It is a real pleasure to take part in the open source ecosystem, and threads like this one are the reason.

(@Dev-next-gen)

PR #597, opened 2026-09-17
test(security): a wrapped answer is still an answer

Dev-next-gen:
## What does this PR do?

This is the wrapped-`json(` follow-up from #596. The second rule in `no-raw-exception-to-the-caller` now reads each answer from `json(` or `new Response(` up to the parenthesis that closes it, however many lines that takes, and names the line the `String(` is on. It no longer needs both on one line.

I bounded it by the call's own parentheses rather than by a window of N lines because I measured the window first and it doesn't leave any room. Two lines miss your three-line `json({` form. Three lines already flag `dictate.ts:172`, where a `new Response(child.stderr)` sits two lines above a catch that isn't part of it. Five lines flag 5 places, ten flag 10. The parentheses give no false alarms: across the 812 answers in `server/src`, a plain paren count closes every one at the same place as a scan that skips strings, templates and comments, and none of the 58 answers that span several lines contains a caught exception's text.

As you asked, the limits that remain are written next to the `e.message` allowance. Parentheses are counted, not parsed, so a lone `)` inside a string in an answer ends it early. And a value built before the call (a variable, or a helper's return) isn't followed. I measured both, and both pass silently, which is why they're written down.

## How was it tested?

I put each leak back one at a time and ran `cd server && bun test test/no-raw-exception-to-the-caller.test.ts`:

| leak put back | `main` | this branch |
|---|---|---|
| none | 3 pass, 0 fail | 3 pass, 0 fail |
| original `index.ts:6717` | 1 fail, `index.ts:6717` | 1 fail, `index.ts:6717` |
| original `index.ts:7614` | 1 fail, `index.ts:7614` | 1 fail, `index.ts:7614` |
| original `prs.ts:3190` | 1 fail, `prs.ts:3190` | 1 fail, `prs.ts:3190` |
| `/view/image` as `json({` / `ok` / `error: ...${String(e)}` / `}, 500)` | 3 pass, 0 fail | 1 fail, `index.ts:6719` |
| `/view/image` as `json(` / `{ ok, error }` / `500,` / `)` | 3 pass, 0 fail | 1 fail, `index.ts:6718` |
| `/walkthrough` with `error: String(e?.message \|\| e),` on its own line | 3 pass, 0 fail | 1 fail, `index.ts:7618` |
| `prAsset` as `new Response(` / `` `upstream: ${String(e)}`, `` / `{ status: 502 },` | 3 pass, 0 fail | 1 fail, `prs.ts:3191` |
| `note: "sorry :)"` before the leak (documented limit) | 3 pass, 0 fail | 3 pass, 0 fail |
| leak in a `const` first, then `json({ error: said })` (documented limit) | 3 pass, 0 fail | 3 pass, 0 fail |

I also ran `oxlint` on the file (0 warnings, 0 errors), `cd server && bun run typecheck` (clean) and the full `cd server && bun test --timeout 20000`: 5327 pass, 19 skip, 1 fail. The failure is `gh-lookup-retries`, which needs a `gh` on PATH. The machine I ran it on has no `gh`, and that file fails the same way on `main` there (1 fail, 4 tests). Only the test file changes, so nothing in `src/`, `web/` or the contract moves.

Found by a defect-hunting pipeline I build and run ([Dev-next-gen](https://github.com/Dev-next-gen)), using Claude Code with Anthropic's Claude Opus 5.


SirAllap:
Thank you — you said you'd take it and it was open inside half an hour, with the measurements done. Merged as `fba5dfb`.

I ran it here rather than take the table on trust, and it holds:

| | |
|---|---|
| the wrapped leak I showed you | caught, at `index.ts:6719` — the line of the `String(`, not of the `json(` around it |
| the three one-line leaks from #596 | still caught, same lines: `6717`, `7614`, `3190` |
| clean tree | `3 pass, 0 fail` — no false alarms across `server/src` |

Two things I want to name, because they are the reason this was easy to accept.

**You ruled out the easy fix with numbers.** A window of N lines is what most people reach for, and "two misses the plainest wrapped form, three already flags `dictate.ts` where a `catch` follows a `new Response(child.stderr)`, five flags 5, ten flags 10" is the difference between a decision and a preference. Counting the call's own parentheses falls out of that, and reads as obvious only *after* the measurement.

**Your two stated limits are true.** I checked both rather than assume:

- a `)` inside a string in an answer ends the scan early and a leak after it goes unseen — confirmed, `0 fail`;
- a `String(e)` put in a variable first and answered with later is not followed — confirmed, `0 fail`.

Writing down the holes in what you just handed over is rare, and it is exactly what makes the next person able to tell a ceiling that was chosen from one that was missed. You also drew the right line on the second: following a value to where it is sent is data flow, and that is CodeQL's job, not a regex's.

One cosmetic thing, not worth a PR on its own: the new sentence about the lone `(` runs a good deal longer than the lines around it in that comment block. Worth a wrap next time anything else in the file moves.

Two merged in a morning, and the second one closed the gap the first one left. Good to have you here.

---

Dev-next-gen:
@SirAllap Thank you for this. Both kinds of remark carry real weight for how this pipeline develops, the approving ones and the critical ones alike, and a reviewer who re-runs the table instead of taking it on trust teaches me more than a green check ever does. Confirming the two limits I wrote down, rather than assuming they were accurate, is the part I did not expect.

Noted on the long line in that comment block. It will get wrapped the next time anything else in that file moves, rather than on its own.

It is a pleasure to contribute to open source worldwide, and to this project in particular.

(@Dev-next-gen)

